PR #112, opened 2019-02-25
MessageLogger interface; deprecate AdaptrisMessage#toString(boolean)

quotidian-ennui:
From [INTERLOK-2647](https://adaptris.atlassian.net/browse/INTERLOK-2647)

* Adds MessageLogger Interface with ADR : https://github.com/adaptris/interlok/blob/feature/INTERLOK-2647/docs/adr/0001-add-messagelogger-interface.md
* Marks AdaptrisMessage#toString(boolean) as deprecated
* Marks AdaptrisMessage#toString(boolean, boolean) as deprecated
* Adds various MessageLogger implementations.
* LogMessageService should support MessageLogger
* Remove use of toString(boolean) in interlok-core (WorkflowImp, FailedMessageRetrier, OnMessageHandler etc etc).


quotidian-ennui:
I'm thinking that if we're changing a "core interface" -> which is what we've done here, we should raise an ADR on develop, and then immediately branch onto feature/XXX and then review before merge.

higgyfella:
the ADR is now: https://github.com/adaptris/interlok/blob/develop/docs/adr/0001-add-messagelogger-interface.md

PR #119, opened 2019-03-01
INTERLOK-2668 - Add Cache to SchemaValidator

quotidian-ennui:
- Schema is not created until first use, and then cached which means we needed to change some of the tests.
- Deprecate schema-metadata-key and make schema %message{}
- Add protection for if people have configured c:/x/y/z as their schema via FsHelper (it wouldn't have worked anyway, but now it could...)


quotidian-ennui:
@aaron-mcgrath-adp  -> I've fixed the javadocs; does it now make more sense ;)

PR #121, opened 2019-03-04
Feature/interlok 2624

aaron-mcgrath-adp:
A new JMS Async producer.
Doesn't really care about the jms bridge'ing consumer ack issues.
Just allows you to configure an exception handler when a message fails.

Also included is the beginnings of a configurable consumer ack handler, with the idea being that we will need it if we're worried about message loss when using async producers.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/jms/AcknowledgementHandler.java
+++ b/interlok-core/src/main/java/com/adaptris/core/jms/AcknowledgementHandler.java
@@ -1,0 +1,12 @@
+package com.adaptris.core.jms;
+
+import javax.jms.JMSException;
+import javax.jms.Message;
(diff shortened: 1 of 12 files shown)


quotidian-ennui:
@aaron-mcgrath-adp 
Test-Class for async-producer doesn't actually test the lifecycle ever (so we have 0% coverage on my newly added init() method for null production)

This does "break the workflow" in a way; if you end up firing the StandardProcessingExceptionHandler and it writes to filesystem. You have a copy of the message that was attempted to be sent to the JMS Broker (not the original) -> so you can't retry the message, even if you knew which workflow it came from.

That isn't necessarily a problem; it's just something that probably should be documented in the class javadocs.

PR #123, opened 2019-03-07
Add default implementation to ComponentLifecycle

quotidian-ennui:
https://github.com/adaptris/interlok/blob/feature/ADR-0002/docs/adr/0002-add-default-implementations-to-componentlifecycle.md

* Changes already applied to ComponentLifeCycle (it's a single file change)
* ./gradlew check works.


quotidian-ennui:
ADR merged back into develop : https://github.com/adaptris/interlok/blob/develop/docs/adr/0002-add-default-implementations-to-componentlifecycle.md

PR #134, opened 2019-03-21
INTERLOK-2702 Merge functionality from DynamicServiceLocator into DynamicServiceExecutor

quotidian-ennui:
* Deprecate all of DynamicServiceLocator and associated classes (@Removal for 3.11.0)
* Add in new ServiceExtractor implementations to replicate the existing functionality
   * URL (which replicates)
   * Database
   * Cache
   * Any DataInputParameter<String> 

https://github.com/adaptris/interlok/blob/feature/INTERLOK-2702/docs/adr/0003-deprecate-dynamicservicelocator.md

### DIFF
--- a/docs/adr/0003-deprecate-dynamicservicelocator.md
+++ b/docs/adr/0003-deprecate-dynamicservicelocator.md
@@ -1,19 +1,36 @@
-# 3. Deprecate DynamicServiceLocator
+---
+layout: page
+title: 0003-deprecate-dynamicservicelocator
+---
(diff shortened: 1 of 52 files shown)


quotidian-ennui:
Well, would it ever be a pointer to an actual object service-list -> the implication of that is that the service might have to be threadsafe etc (which is certainly an option, since we have shared-service#deepClone)

If we change the ServiceExtractor return the service; that would probably be fine, it's worth thinking about, It would only mean deprecating the marshaller config in DSE.

Since we might want to; I'll remove the functionalInterface from ServiceExtractor...

PR #137, opened 2019-03-22
INTERLOK-2707 Modify ServiceExtractor Interface to return a Service

quotidian-ennui:
- Added a `getService(AdaptrisMessage, AdaptrisMarshaller)` method; it is a default interface method with appropriate javadocs saying how you should just extend ServiceExtractorImpl if you can't be bothered
to change your impl.
- Added ServiceExtractorImpl
- DefaultServiceExtractor doesn't extend it, so we get coverage...
- Switch DynamicServiceExecutor to use the new method.

Tests still pass with no changes (and still give 100% 👍 ).


interlok-core/src/main/java/com/adaptris/core/services/dynamic/DefaultServiceExtractor.java
```diff
@@ -27,11 +27,11 @@
  * Default {@link ServiceExtractor} implementation that treats the entire payload as the service.
```
quotidian-ennui: Action card entry.

PR #138, opened 2019-03-22
INTERLOK-2703 Add copy-metadata flag

quotidian-ennui:
- Add copy-metadata that defaults to false for backwards compat If true, then copies the original metadata into the child message; never copies object metadata, since the thing you do in the splitter is to produce it somewhere (which means you lose the object metadata)
- Made MaxRowsPerMessage an Integer
- Added NotNull messages for INTERLOK-2705


quotidian-ennui:
Yeah, we never copied it previously, cos Gerco wrote it for an agility GK thing and never needed to, since no metadata was required (all required data was in the DB).

PR #145, opened 2019-04-02
INTERLOK-2725 Add MetadataLogger interface

quotidian-ennui:
- Add MetadataLogger interface
- Add MetadataServiceImpl that provides logging behaviour
- Changed services to extend MetadataServiceImpl and use new logging behaviour.
- Remove redundant methods provided by MetadataServiceImpl
- Add 2 implementations TruncateMetadata + MetadataKeysOnly; the default impl is just a lambda.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/MessageLoggerImpl.java
+++ b/interlok-core/src/main/java/com/adaptris/core/MessageLoggerImpl.java
@@ -16,7 +16,6 @@
 package com.adaptris.core;
 
 import java.util.Collection;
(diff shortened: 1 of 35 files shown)


quotidian-ennui:
```
logMetadata(LogWrapper, String, Collection<MetadataElement>)
logMetadata(LogWrapper, String, Collection<MetadataElement>)
```

LogWrapper is a functional interface, which has the standard TRACE/DEBUG/INFO/WARN/ERROR levels defined as static lambdas.

PR #191, opened 2019-06-14
INTERLOK-2834 Add MultipartMessageBuilder

quotidian-ennui:
Added supporting ByteArray message wrapper implementations.
Add MultipartMessageBuilder which supports
- Arbitrary headers at the top level via mime-header-filter
- Custom expression based Content-ID

Nested parts are created using inline-mime-part-builder (working off byte-arrays) these support
- Expression based Content-ID + Content-Type + Content-Transfer-Encoding
- Arbitrary headers built from metadata for each part.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/DefaultAdaptrisMessageImp.java
+++ b/interlok-core/src/main/java/com/adaptris/core/DefaultAdaptrisMessageImp.java
@@ -17,7 +17,6 @@
 package com.adaptris.core;
 
 import static org.apache.commons.lang.StringUtils.isEmpty;
-
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.FilterOutputStream;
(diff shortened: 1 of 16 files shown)


quotidian-ennui:
```
  <multipart-message-builder>
   <unique-id>efd6c85f-5776-4e60-ac41-2937c46e26bd</unique-id>
   <inline-mime-body-part-builder>
    <body class="byte-array-from-object-metadata">
     <key>ObjectMetadataKey</key>
    </body>
    <content-type>application/octet-stream</content-type>
    <content-id>from-object-metadata</content-id>
   </inline-mime-body-part-builder>
   <inline-mime-body-part-builder>
    <body class="byte-array-from-metadata">
     <key>StringMetadataKey</key>
    </body>
    <content-type>application/octet-stream</content-type>
    <content-id>from-string-metadata</content-id>
   </inline-mime-body-part-builder>
   <inline-mime-body-part-builder>
    <body class="byte-array-from-payload"/>
    <part-header-filter class="regex-metadata-filter">
     <include-pattern>Content-Disposition</include-pattern>
    </part-header-filter>
    <content-id>%message{%uniqueId}</content-id>
   </inline-mime-body-part-builder>
  </multipart-message-builder>
```

PR #214, opened 2019-07-29
New conditionals

bklair:
New conditional classes for Agx work: not condition and the isEmpty operator


bklair:
Agreed to comments, I also find a minor bug in the isEmpty code, code coverage now 100% on my local PC for the changes

PR #215, opened 2019-07-30
ADR-0005 - Deprecate and remove Destination

quotidian-ennui:
It's a big change, so we need to think about the consequences.
https://github.com/adaptris/interlok/blob/adr/ADR-0005/docs/adr/0005-remove-produce-destination.md

(or after merge : https://github.com/adaptris/interlok/blob/develop/docs/adr/0005-remove-produce-destination.md)


quotidian-ennui:
I think the biggest issue is the fundamental change to the produce() / request() methods defined in the AdaptrisMessageSender / AdaptrisMessageProducer interfaces.

In the main trunk at least; we don't usually use the produce(msg, dest) method (it's usually just called by produce(msg))

I'm not sure what the impact is of marking the methods with ProduceDestination in them as deprecated.

PR #217, opened 2019-07-31
INTERLOK-2852 Add a condition to MessageAggregator

quotidian-ennui:
- Add a default method to MessageAggregator
- Add a Conditon to AggregatorConsumerImpl / SplitJoin.
- JsonArrayAggregator still needs to override the default method, since it has the additional "include-failed-evaluations" capability.


quotidian-ennui:
Because of https://github.com/adaptris/interlok/pull/219 this is a defunct PR.

PR #219, opened 2019-08-02
INTERLOK-2852 / AGX Filter splitter3

bklair:
This pull request is requires the earlier pull request: new conditional classes.
This functionality is the introduction of a filter to the aggregator classes: Interlok 2852

The following are the core aggregators:
XmlDocumentAggregator
IgnoreOriginalXmlDocumentAggregator
ZipAggregator
MimeAggregator
IgnoreOriginalMimeAggregator
ReplaceWithFirstMessage

The following are the json project aggregators which are in a separate pull request(interlok-json)
JsonArrayAggregator
JsonArrayArrayAggregator

Tests have been written to ensure all aggregators can actually filter their msgs. I've also added a test to handle a conditional throwing a core exception and testing that the retain msgs flag retains such msg in the output.

There may be other aggregators in other project, all that needs to be done is to add the filter method to the for loop within the join method.

Additionally the docs will need to be updated to detail the new functionality.


quotidian-ennui:
@bklair I raised a PR that did the same thing in a slightly different way... https://github.com/adaptris/interlok/pull/217

Both are acceptable; so please choose; one way means that aggregators need to know about filters (and executing it); the other way means that things that call aggregators have to know about filters and filter it.

bklair:
I've gone over both implementations, it all comes down to where the filter/condition is added, is it logical to have it in the Aggregator? Certainly the aggregator is it's own self contained service and can be dropped into other components unlike the SplitJoin. Lewin's implementation is just missing the retain messages that had exceptions thrown in the condition check. 
Over to @aaron-mcgrath-adp for the final say.

PR #231, opened 2019-08-28
INTERLOK-2462 Bump validator to 6.0.17

quotidian-ennui:
- Bump hibernate validator to 6.0.17
- Bump validation-api to 2.0.1
- Bump javax.servlet-api to 4.0.1
- Move oro:oro into testCompile since it has been removed as a regexp provider since 3.9.0 (we still use it in tests).
- Remove commons-pool:1.6 (it might still get pulled in but we don't depend on it).
- Remove dependency on javax.el:javax.el-api
- Add dependency on org.glassfish:javax.el (since this is the full ExpressionFactory impl)
- Force exclusion on javax.el-api


quotidian-ennui:
Tested @ runtime with existing interlok-core nightly via ant+ivy with validateConfig=true in bootstrap.properties

```
      <dependency org="javax.validation" name="validation-api" rev="2.0.1.Final" conf="runtime->default"/>
      <dependency org="javax.servlet" name="javax.servlet-api" rev="4.0.1" conf="runtime->default"/>
      <dependency org="org.hibernate.validator" name="hibernate-validator"  rev="6.0.17.Final" conf="runtime->default"/>
      <dependency org="org.glassfish" name="javax.el" rev="3.0.1-b11" conf="runtime->default"/>

      <exclude org="javax.el" artifact="javax.el-api"/>
      <exclude org="org.hibernate" artifact="hibernate-validator"/>
```

Interlok starts up, and the UI starts (but not tested UI interactions).

quotidian-ennui:
As per discussion on teams we should write new tests for checking constraint violations on 
* ChannelList
* WorkflowList
* ServiceCollectionImp

and remove @Valid on the underlying list that they wrap (since they implement Collection themselves anyway).

PR #254, opened 2019-09-27
INTERLOK-623 review advanced config

quotidian-ennui:
see the checkin comments; tried to make each one obvious as to what it affects since it's quite a lot.

Basically for those things already annotated with AdvancedConfig; review them to see if they should be annotated with the `rare=true` as well.


quotidian-ennui:
@higgyfella  @sebastien-belin-adp  -> please could you review how these new things display in the settings editor

PR #265, opened 2019-10-08
INTERLOK-2860 Add RemoteBlobFilter and implementation

quotidian-ennui:
- Add RemoteBlobFilter and RemoteBlobFilterWrapper which wraps a FileFilter (for easy filtering).
- Moved com.adaptris.core.util.Args into com.adaptris.interlok.util however, not deprecating it because of excessive use, but we have to consider what happens in 4.0 since interlok-common probably shouldn't exist as a sub-project.


quotidian-ennui:
Raised the class name thing as INTERLOK-2997

PR #267, opened 2019-10-10
Allow adding workflow, channel, messageid and any other message data to MDC

gerco:
This can be used in combination with an appropriate logging Appender to log message id and such to ElasticSearch for better tracing of a message.


quotidian-ennui:
@gerco happy to approve it; but do you mind raising a ticket in Jira around your use-case (and possibly example config) since I think you've changed the semantics of poolingworkflow slightly (don't think it's a problem; but we might want to think about side-effects).

PR #289, opened 2019-11-05
ADR-0006 Add a callback to AdaptrisMessageListener.

quotidian-ennui:
https://github.com/adaptris/interlok/blob/adr/ADR-0006-workflow-callback/docs/adr/0006-workflow-callback.md

Once merged then it is : https://github.com/adaptris/interlok/blob/develop/docs/adr/0006-workflow-callback.md (until merged, this just has dummy content).

### DIFF
--- a/docs/adr/0006-workflow-callback.md
+++ b/docs/adr/0006-workflow-callback.md
@@ -1,76 +1,86 @@
 ---
 layout: page
-title: template
+title: 0006-workflow-callback
 ---
-# [short title of solved problem and solution]
+# Add a new method to AdaptrisMessageListener
 
-* Status: [accepted , superseeded by [ADR-0005](template.md) , deprecated , ...] <!-- optional -->
-* Deciders: [list everyone involved in the decision] <!-- optional -->
-* Date: [YYYY-MM-DD when the decision was last updated] <!-- optional -->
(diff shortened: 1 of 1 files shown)


gerco:
Do we need a new Functional Interface for this? Is this something we could do with `Consumer<AdaptrisMessage>`? That would save defining an extra interface and still work the same with Lambda's, etc.

quotidian-ennui:
> Do we need a new Functional Interface for this? Is this something we could do with `Consumer<AdaptrisMessage>`? That would save defining an extra interface and still work the same with Lambda's, etc.

You're almost certainly right; we should probably re-use the consumer interface; I shall update the doc; however, other than that, no objections to the context and intention @gerco

gerco:
> > Do we need a new Functional Interface for this? Is this something we could do with `Consumer<AdaptrisMessage>`? That would save defining an extra interface and still work the same with Lambda's, etc.
> 
> You're almost certainly right; we should probably re-use the consumer interface; I shall update the doc; however, other than that, no objections to the context and intention @gerco

No objections. This may actually be a more general way to handle the Solace workflow @aaron-mcgrath-adp  is currently working on since it may allow for things like async acks in JCSMP as well.

PR #292, opened 2019-11-06
Fix typo for deleteAggregatedFiles in @DisplayOrder

sebastien-belin-adp:
Just fixed a typo for the property deleteAggregatedFiles in @DisplayOrder annotation.


sebastien-belin-adp:
appveyor fails but I'm not really sure why as I just fixed a typo.

quotidian-ennui:
Appveyor fails becuase it's networking is crap and/or 
-> it's why we have 3 CI builders.... if all 3 fail, then we know it's bad, 1, then we can safely ignore it (!)

PR #297, opened 2019-11-20
INTERLOK-2997 #comment  Add  implemts to InputFieldHint annotation.

sebastien-belin-adp:
It will helps displaying a list of suggestions for the field value in the
UI.


sebastien-belin-adp:
Use the new annotation attribute for the fileFilterImp fiedl in FsConsumerImpl and RemoteBlobFilterWrapper.
I'm sure we can find more use for it later.

sebastien-belin-adp:
For some reason AppVeyor failed again on SplitJoinServiceTest.
But I haven't changed anything related to it.

sebastien-belin-adp:
I was talking with Aaron about naming and we thought maybe about implementation but I thought that was a bit long so I chose implemts but ofType is working fine for me.

PR #298, opened 2019-11-20
INTERLOK-3003 Add support for primitives in FileFilter constructors

quotidian-ennui:
- Add FileFilterBuilder to create the FileFilter instances.
- Switch FsHelper to use the new builder.


quotidian-ennui:
> 
> 
> What happened to the builder methods withFilterExpression and withFilterImpl?

Probably a badly named builder (happy to change the name if you prefer) since the actual build method is a static method.

Will make it non static actually, `new FileFilterbuilder().withX().withY().build`; no reason not (as the methods don't have to be static).

PR #299, opened 2019-11-20
INTERLOK-3080 Extended Logging Context

mcwarman:
- LoggingContextWorkflowInterceptor adds object metadata
- AddExtendedLoggingContext optionally adds metadata based on that
- RemoveLoggingContext switched to using list


mcwarman:
Makes sense, something like this?

gerco:
Looks good to me

PR #300, opened 2019-11-20
Github action for opensource components.

quotidian-ennui:
gradle build on push; github action.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -20,6 +20,20 @@ jobs:
         sudo apt-get -y install haveged
         sudo systemctl enable haveged
         sudo systemctl restart haveged
+    - name: Gradle Wrapper Cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.gradle/wrapper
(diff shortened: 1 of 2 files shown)


quotidian-ennui:
Thoughts? Still quite fast, even w/o a cache.
Marked JMS tests as skipped; enabled haveged.

gerco:
As long as this doesn't need to access our private Nexus for dependencies, sounds good. We won't be able to move a lot of our optional components to Github CI due do dependencies and them not being open source but that doesn't have to stop us from doing this to core (and a few others?)

quotidian-ennui:
> 
> 
> As long as this doesn't need to access our private Nexus for dependencies, sounds good. We won't be able to move a lot of our optional components to Github CI due do dependencies and them not being open source but that doesn't have to stop us from doing this to core (and a few others?)

We can have secrets apparently (not sure how they get setup); but since nothing in github is closed source (i.e. nothing has references to private nexus); this is fine. Everything will eventually also build with a gitlab runner as well :)

PR #301, opened 2019-11-21
Feature/INTERLOK-3051

albinoloverats:
Allow multi-payload messages to be MIME encoded/serialized for storage and retrieval.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/MimeEncoderImpl.java
+++ b/interlok-core/src/main/java/com/adaptris/core/MimeEncoderImpl.java
@@ -97,6 +97,29 @@ public abstract class MimeEncoderImpl extends AdaptrisMessageEncoderImp {
     }
   }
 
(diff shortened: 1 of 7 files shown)


quotidian-ennui:
@albinoloverats -> I would have expected the MimeEncoder (the actual MimeEncoder) would have the same changes as you made for FileBackedMimeEncoder...
At the moment if you configure a MimeEncoder (rather than a FileBackedOne); it won't support MultiPayloads ...

quotidian-ennui:
@albinoloverats -> MultipartMimeEncoder needs to extend MimeEncder and work on "outputstream/inputstreams" rather than Files..  
Otherwise you won't just "work" with non-filesystem based producers and consumers; or infact with EncodingService...

c.f. AdaptrisMessageWorkerImp#encode/decode...
c.f. com.adaptris.core.services.transcoding.EncodingService 

-> We probably need to have some kind of "generics" on MME / ME , and the MessageEncoder interface. stupid legacy 1.4

PR #328, opened 2020-01-07
feature/INTERLOK-3087

albinoloverats:
Multi-payload message for-each.

A multi-payload message is split into many messages, one for each of its payloads and then loop through each payload and perform the desired service (list) on each one.


albinoloverats:
@quotidian-ennui I just had it match the class name. Though I guess it's easier to rename the class than have users update their config if more for-each's are ever added.

PR #333, opened 2020-01-08
INTERLOK-2867 Remove GenerateBeanInfo annotation

quotidian-ennui:
- Remove GenerateBeanInfo as an annotation and associated Processors.
- Tidy up final use from commons-lang2
- Since we have removed commons-lang2 from the dependency chain; this may break some downstream optional builds (since commons-lang2 will no longer be available); they will have to physically include it as a dependency.


quotidian-ennui:
WIll break interlok-drools (but that's a deprecated project...)

PR #349, opened 2020-01-24
feature/INTERLOK-2761

albinoloverats:
* Update ExternalResolver class to provide `resolve(String, String)` method
* Add default `resolve(String, String)` method to Resolver interface
* Add new exception type for resolve related issues
* Add new resolver for payloads using XPath
* Add reference to new resolver for service loader
* Add and update unit tests as necessary


quotidian-ennui:
* Could you make UnresolveableMetadataException extend UnresolvableException? (this is the exception that's thrown when the metadata key doesn't exist during resolve()

* Since we aren't allowing the user to "configure" the document builder; please use DocumentBuilderFactoryBuilder#newRestrictedInstance() instead of creating a DocumentBuilderFactory directly -> because XXE !!!

* What troubles me is that if we have something that's configured like this
```
key=%payload{xpath:/xyz}%payload{xpath:/abc}
```
Then it will "create" a  DOM + xpath for *every resolution* -> so you probably want to create the DOM once on the String if you have a match, outside the loop, otherwise the performance is going to suck badly very quickly.

quotidian-ennui:
I didn't spot it last time... 
```
XPathExpression expr = XPathFactory.newInstance().newXPath().compile(path);
```

Use com.adaptris.util.text.xml.Xpath.newXpathInstance() instead; otherwise even if we wanted to support XPath 2 features, we can't... now that you've moved the class into interlok-core; it's probably not a big deal...

(read the docs if you fancy understanding "opinions").

quotidian-ennui:
Like Columbo, there's always a "finally"...  It all looks good, until I spotted that you are using String.getBytes() to turn it into a ByteArrayInputStream...

Would it be better to change "target" in the `resolve(String, String)` method to be `resolve(String, InterlokMessage)` so that we can use InterlokMessage#getInputStream()  ..,.; this wouldn't necessarily break the JSON equivalent since JsonPath can work with inputStreams...

This would have the handy benefit that we isolate ourselves from String.getBytes() and other encoding issues since these are always the issues with our French customers...

What I'm thinking of is that in the specific XML case; it's always "best practise to let the XML Parser decide what the encoding is, based on the `xml version=1.0 charset=` directive. So if we have something like this

* AdaptrisMessage that has a encoding of 'null' (i.e. the platform default, cp1252 on Windows...)
* An XML document that has `<?xml version="1.0" charset="ISO-8859-2"?>` at the start.

The simple act of doing String#getBytes() in the resolve method means that we will actually have duff data since the byte representation of ISO8859-2 is different to cp1252...

PR #353, opened 2020-01-29
feature/interlok-3141

albinoloverats:
Add implementations of `DataInputParameter` + `DataOutputParameter` that are multi-payload aware so that:

1. `MultiPayloadStreamInputParameter` -> that gives you the named "payload" as an input stream
2. `MutiPayloadStringInputParameter` -> as above but as a `String`
3. `MultiPayloadByteArrayInputParameter` -> as above etc etc

and:

1. `MultiPayloadStreamOutputParameter` -> creates a new part if the part doesn't exist; and returns you an output stream to write to it.
2. `MultiPayloadStringOutputParameter` -> blah but as a `String`
3. `MultiPayloadByteArrayOutputParameter` -> blah but as a `byte[]`

If the message type isn't an `MultiPayloadAdaptrisMessage` we just throw an exception.

Possible improvements *(suggestions welcome)* before merge include considering how the naming of the multi-parts works... and where that is stored.


quotidian-ennui:
The names are fine I think

I'm thinking that we need to make the payload-id's "resolvable" from metadata... which means we should also change the add-payload / remove-payload / switch-payload services as well...

i.e. we do this instead

```
  @InputFieldDefault(resolvable=true)
  private payloadId;
  public byte[] extract(InterlokMessage m) throws InterlokException  {
    if (m instanceof MultiPayloadAdaptrisMessage)   {
      return extract(m.resolve(getPayloadId()), (MultiPayloadAdaptrisMessage)m);
    }
    throw new InterlokException("Cannot extract payload from message type " + m.getClass().getName() + " as it does not support multiple payloads.");
  }
```

That has a possibly amusing side-effect of someone perhaps at some point doing a payloadId of `%multipayload{all_the_payload_ids:json:$.source}` in MultipayloadInputStreamDataInputParameter...

Does the interface MultiPayloadDataInput / MultiPayloadDataOutput need to exist; I think the getters & setters defined in there definitely don't; You aren't physically using the interfaces anywhere directly; having the interface define getters and setters is a bit dangerous anyway (cos naming!).

Would changing the InputParameter/OutputParameter interface help I wonder?

```
public interface DataInputParameter<T, E extends InterlokMessage> {
  T extract(E m) throws InterlokException;
}
```

- I'm not sure that it would since generics are only a compile time check I think; getting a ClassCastException at runtime wouldn't be that useful for support...

PR #366, opened 2020-02-14
INTERLOK-3170 Switch to using java.util.Base64

quotidian-ennui:
- Add EncodingHelper that wraps the various MimeUtility / Base64 wrappers.
- EncodingHelper also wraps an enum for the Base64 style
- Deprecate Conversion#base64 operations
- Base64ByteTranslator now uses java.util.Base64
- MetadataToPayload / PayloadToMetadata switched to use EncodingHelper.Encoding enum
- Base64Encode / Decode / EncodeMetadata / DecodeMetadata now uses one of the Base64 styles.

It's arguable that Conversion#base64 methods should just delegate to Base64.getEncoder/getDecoder so happy to do that to avoid lots of deprecation warnings...

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/Base64DecodeService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/Base64DecodeService.java
@@ -16,16 +16,13 @@
 
 package com.adaptris.core.services;
 
-import javax.mail.internet.MimeUtility;
-
+import java.io.InputStream;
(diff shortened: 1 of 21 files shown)


quotidian-ennui:
@mcwarman  -> this should help with the Salesforce thing.

quotidian-ennui:
commons-codec -> still does it as "MIME style" from what I could work out; so the java.util Base64 is the right thing to use.

PR #381, opened 2020-03-05
Update ForEach.java

higgyfella:
## Motivation

because the since value is wrong, in the UI, we have 2 lists, a 3.10 list and a 3.10.0 list, and since 3.10.0 is the correct value, this class should be updated

## Modification

I changed the since value in the component profile annotation

## Result

that in the ui, the components sidebar, there will be only 1 group 'since 3 10 0' (with 3 components)

## Testing

if you using the ui, open config page, open config sidebar, goto components panel, and using the groups, you'll see 'since <value>', there should only be 1 entry for the 3.10.0 release


quotidian-ennui:
@higgyfella  -> does this mean we need a B4 ;)

PR #392, opened 2020-03-25
INTERLOK-3124 Added Management Components to JMX

aaron-mcgrath-adp:
## Motivation

We can configure management components in the bootstrap.properties, but then they are forgotten.  Almost like mini programs that run in the background.  What if we later want to know what management components are running?  Or even if they all started up correctly?

## Modification

Firstly I have had to modify the ManagementComponentFactory such that it returns a list of information on running management components; a new class ManagementComponentInfo.

I've also taken the time to make the ManagementComponentFactory less rubbish, by  removing all the reflection.

## Result

The AdapterRegistry exposes a bunch of JMX endpoints, it also now exposes a list of management components, including their current state; init, start, stop and closed.

## Testing

You'll need to start the adapter with at least one management component, then simply check jconsole for the AdapterRegistry, which should now give you a getManagementComponentInfo().
It's a list of ManagementComponentInfo's.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/management/ManagementComponentFactory.java
+++ b/interlok-core/src/main/java/com/adaptris/core/management/ManagementComponentFactory.java
@@ -68,7 +68,7 @@ public class ManagementComponentFactory {
       try {
         manCompInfo.getInstance().init(bootstrapProperties);
       } catch (Exception e) {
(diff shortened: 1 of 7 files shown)


aaron-mcgrath-adp:
> I wonder why it was done via reflections in the first place (I did it).

I think it was me and Ash a few years ago when we added classloaders into management components.  No idea why we chose to do it that way?

PR #402, opened 2020-03-31
INTERLOK-3214 New Metadata service Escape Single and Escape Double quotes

Co-Din:
## Motivation

Why am I doing this

New services for escaping the single and double quote characters in metadata values were requested.

## Modification

What did I change

I created two new services in the core-services-metadata which escape the single and double quote characters adding a backslash before them.
Complimenting tests were also added for these two new services.

## Result

The user may now add a service for escaping their metadata's values of single or double quote characters.

## Testing

There are two ways to test. The first would be to run the added Junit tests and the second to build the core with the new services and drop the jars into a working instance of an Interlok adapter and add the service and use them.


Co-Din:
> 
> 
> I would probably change the tags to be just "service,metadata"
> Also add a since="3.10.1" to the component profile.

Just updated, thanks for pointing out.

PR #416, opened 2020-04-21
INTERLOK-3225 auto cache oauth token

quotidian-ennui:
## Motivation

In almost all cases people want to get an OAUTH Access Token and re-use it repeatedly which means that they invariable have something like this as a shared-service (or any variation thereof).

```
<if>
  <exists-in-cache/>
  <then>
    <get-single-value-from-cache/>
  </then>
  <otherwise>
    <get-oauth-token/>
    <add-single-value-to-cache/>
 </otherwise>
</if>
```

Since this is a often used pattern (and in fact probably only use-case) then we should shorten the chain so it's not so unwieldy.

## Modification

* Refactor the existing add-single-value-to-cache so that the cache-expiry logic is exposed via a helper method in c.a.c.c
* Refactor get-oauth-token so that we now encapsulate what we do with the tokean via a `access-token-writer` interface (the only impl is `oauth-access-token-to-metadata`).
* Add a get-and-cache-oauth-token service that does everything...

## Result

A new service which can be configured, that allows you specify the underlying cache. Note that if an expiry exists then it will be used, so the underlying cache instance needs to support per-item-expiry otherwise an UnsupportedOperationException will be thrown (ExpiringMapCache is fine).

```
  <get-and-cache-oauth-token>
   <access-token-writer class="oauth-access-token-to-metadata">
    <token-key>Authorization</token-key>
   </access-token-writer>
   <access-token-builder class="com.adaptris.core.http.oauth.DummyAccessTokenBuilder"/>
   <cache-key>%message{cachekey}</cache-key>
   <connection class="cache-connection">
    // Defaults to expiry-map-cache with sensible defaults.
   </connection>
  </get-and-cache-oauth-token>
```

## Testing

You'll need something that requires OAUTH (microsoft graph ??); use get-and-cache-oauth-token where you would have used get-oauth-token. Do the operation > 1 time, and you should see logging about it being "cached".

```
            <services>
              <get-and-cache-oauth-token>
                <unique-id>get-oauth-token</unique-id>
                <access-token-writer class="oauth-access-token-to-metadata">
                  <token-key>Authorization</token-key>
                </access-token-writer>
                <access-token-builder class="azure-username-password-access-token">
                  <client-id>${azure.clientid}</client-id>
                  <resource>${azure.resource}</resource>
                  <authority-url>${azure.authority.url}</authority-url>
                  <username>${azure.username}</username>
                  <password>${azure.password}</password>
                </access-token-builder>
                <cache-key>authCache</cache-key>
                  <lookup-name>oauth-cache</lookup-name>
                </connection>
              </get-and-cache-oauth-token>
              <apache-http-request-service>
                <unique-id>MS-graph-ME</unique-id>
                <url>https://graph.microsoft.com/v1.0/me/</url>
                <content-type>text/plain</content-type>
                <method>GET</method>
                <response-header-handler class="apache-http-response-headers-as-metadata"/>
                <request-header-provider class="apache-http-no-request-headers"/>
                <authenticator class="apache-http-configured-authorization-header">
                  <header-value>%message{Authorization}</header-value>
                </authenticator>
              </apache-http-request-service>
              <jetty-response-service>
                <http-status>%message{adphttpresponse}</http-status>
                <content-type>%message{Content-Type}</content-type>
              </jetty-response-service>
            </services>
```

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/cache/CacheExpiry.java
+++ b/interlok-core/src/main/java/com/adaptris/core/cache/CacheExpiry.java
@@ -1,0 +1,86 @@
+package com.adaptris.core.cache;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import com.adaptris.util.TimeInterval;
+import com.adaptris.util.text.DateFormatUtil;
+
+public class CacheExpiry {
+  
+  private static final List<ExpiryParser> EXPIRY_PARSERS =
+      Collections.unmodifiableList(
+          Arrays.asList(new ExpiryParser[] {
+          (val) -> build(val),
+          (val) -> build(DateFormatUtil.parse(val, null))
+      }));
+  
+
+  private static Optional<Expiry> build(final Date date) {
+    return Optional.ofNullable(date).map(d -> Optional.of(new Expiry(d))).orElse(Optional.empty());
+  }
+
+  private static Optional<Expiry> build(final String millis) {
+    // don't use NumberUtils since we don't want to support a - number...
+    return Optional.ofNullable(millis).filter(t -> t.chars().allMatch(c -> c >= '0' && c <= '9'))
+        .map(t -> Optional.of(new Expiry(Long.parseLong(t)))).orElse(Optional.empty());
+  }
+
+  /**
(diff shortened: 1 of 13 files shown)


quotidian-ennui:
I was "hoping" that Optional.map() etc would shorten the jacoco coverage chain; but no, since it probably generates the same "byte code" it's just the same as a bunch of if statements in the end...

PR #446, opened 2020-05-21
INTERLOK 3048 - Add IsIn, NotIn and IsNotEmpty operators

yashashrirane:
Add IsIn, NotIn and IsNotEmpty operators


mcwarman:
merging locally in order to make some subtle  changes around naming of classes

PR #449, opened 2020-05-26
INTERLOK-3290 Add Extended Schema Validation.

quotidian-ennui:
## Motivation

The existing `xml-schema-validator` implementation throws an exception and only ever logs the first exception. This is because the SAX ErrorHandler implementation that is passed to the validator re-throws all exception; so it never gets past the first exception...

It will actually be useful to capture all schema violations such that we can report on all the exceptions rather than just the first one.

## Modification

- Deprecate XmlSchemaValidator and marked for removal in 3.12.0
- Add new BasicSchemaValidator that logs all violations and errors.
- Add new ExtendedSchemaValidator that has pluggable behaviour
  - Add a simple object model and supporting bean style classes for simple serialisation to XML using XStream.
  - `schema-violations-as-object-metadata` to store the schema violations as standard metadata (marshalled XML)
  - `schema-violations-as-metadata` to store the object as object metadata
  - `schema-replace-payload-with-violations` to overwrite the entire payload with the schema violations (marshalled XML)

This actually brings xml schema validation close to parity with json-schema-validation (other than injecting the schema violations into the payload). It's probably quite obtuse to detect a schema violation and then violate the schema further and inject additional information into the document...

## Result

- No change for existing configuration; since xml-schema-validator is unchanged; however a warning will be logged about deprecation
  - It won't show up in the UI
- If you choose `xml-validation-service` then there are now additional options:

  - "basic-xml-schema-validator" which logs all the violations and throw an exception (similar to the behaviour of xml-schema-validator).
  - "extended-xml-schema-validator" which allows for pluggable behaviour (as described above : metadata/object metadata/ payload).

## Testing

If you use the test document `interlok-core/src/test/resources/transform/schema-input.xml` with the test schema `interlok-core/src/test/resources/transform/schema-simple-alternate.xsd` then it *should fail validation* (since simple-alternate alludes to a missing referenceid attribute).

- Configure with the deprecated xml-schema-validator then a single error log is emitted before ServiceException is thrown
- Configure with basic-xml-schema-validator then 2 error logs are emitted before ServiceException is thrown.
- Configure with extended-xml-schema-validator then you can select the behaviour (it will default to the same behaviour as basic-xml-schema-validator) that you want.


quotidian-ennui:
Tested with
```
<xml-validation-service>
  <validators>
    <extended-xml-schema-validator>
        <schema>${schemaUrl}</schema>
        <schema-violation-handler class="schema-replace-payload-with-violations"/>
    </extended-xml-schema-validator>
  </validators>
</xml-validation-service>
 <jetty-response-service>
   <unique-id>cranky-lamarr</unique-id>
   <http-status>200</http-status>
   <content-type>application/xml</content-type>
</jetty-response-service>
```
And the payload show by postman is : 

```
<schema-violations>
    <schema-violation>
        <message>cvc-maxLength-valid: Value &apos;Ola Nordmann&apos; with length = &apos;12&apos; is not facet-valid with respect to maxLength &apos;8&apos; for type &apos;stringtypelengthchecks&apos;.</message>
        <line-number>8</line-number>
        <column-number>30</column-number>
    </schema-violation>
    <schema-violation>
        <message>cvc-type.3.1.3: The value &apos;Ola Nordmann&apos; of element &apos;name&apos; is not valid.</message>
        <line-number>8</line-number>
        <column-number>30</column-number>
    </schema-violation>
</schema-violations>
```

PR #470, opened 2020-06-19
INTERLOK-3230 Channel Restart Produce Exception Handler

Co-Din:
## Motivation

Why am I doing this

When the channel restarts after it's produce exception it still doesn't think it's back up and running, stopping any messages from being processed.

## Modification

What did I change

Set the availability to true when the Channel is started.

## Result

The channel fully restarts upon the channel produce exception handler.

## Testing

Follow the instructions at Interlok-3230 to recreate, with the current change the channel will restart properly.


aaron-mcgrath-adp:
Seeing as this won't make it into 3.10.2, lets add a unit test to prove the availability is toggled correctly.

PR #479, opened 2020-07-09
INTERLOK-3329 Please go away ConsumeDestination

quotidian-ennui:
## Motivation

As part of the move to V4, and to make the "UI look nice" we need to deprecate ConsumeDestination so that consumers don't have an opaque object to render in the UI that's in the abstract parent.

This is the first phase of that.

## Modification

- Remove get/set Destination from AdaptrisMessageConsumer interface
- Remove destination from AdaptrisMessageConsumerImp
- Added deprecated destination directly to the consumers; this does "duplicate"
code to some degree.
- Added a helper class to ease transition (DestinationHelper).
- Remove use of destination in QuartzPoller
- Refactor JmsReplyToWorkflow, even though it's deprecated it offended me.

Although "ConsumeDestination" is not marked as deprecated; it will be once the INTERLOK-3329 is ready to be closed.

## Result

The tests pass; however, the expectation is that all downstream projects will break that contain consumers (e.g. mail / oftp etc).

## Testing

The tests pass


aaron-mcgrath-adp:
Refactor JmsReplyToWorkflow, even though it's deprecated it offended me.

Laughed.  Hard.

quotidian-ennui:
> There's an awful lot going on here.
> Conceptually it looks fine.
> 
> Some questions raised though;
> 
> Do we need to filter white-space from comma separated lists like the accepted methods for HTTP?

It gets trimmed when we create the actual handlers (in the `handler()` method); since we expect people to actually put in "Post, Get, Head, Delete" (like they would cos who wants to SHOUT).

> Or a new UI element that can build the comma separated list with ticks against all available methods (some annotation that lists acceptable input values)?
> 


> Would it be possible to deprecate destination rather than remove it from the interfaces?
> Maybe add advanced/rare annotations to the destination object just so it doesn't show up in UI by default?
> 

I opted for the pain now to remove it from AMC#get/set destintation. Some "consumers" don't have destinations in the same way anyway (SAP springs to mind); so it was kind of always broken. It's mainly to catch the situations where we do `((AdaptrisMessageConsumer) workflow.getConsumer()).getDestination()` and to try and understand why they "wanted access to the destination"... -> turns out there's no good reason (!)

> Can we just delete Pas and Ptp consumers/producers? Or v4?
> 

I think that's a v4 thing; the only problem is in the event of a JMSReplyTo; in the sense that its string representation may not be RFC6167 parseable... we'd probably want to think about that and how all the special destinations are going to manifest itself. I suspect we're going to end up with a `%message{%JMS_REPLY_TO}` as a special case (or do the same resolve for object metadata).

> Out of interest, why don't you like the keyword "this"?
> Is it auto format on eclipse or...?

That's just eclipse doing it's thing (so autoformat).

PR #487, opened 2020-07-22
Added Serialization to the ManagementComponentInfo bean

higgyfella:
## Motivation

The bean required serialization so we could use it on a ui feature : 
  INTERLOK-3251 - 'UI Dashboard - Add Adapter Management Info to the Information modal'

## Modification

Added Serializable to the ManagementComponentInfo class, and made member 'instance' transient, as there's no point (i think) of this being part of the serialized object.

## Result

people can use the AdapterRegistryMBean.getManagementComponentInfo method (either jconsole or ui)

## Testing

I tested using the UI dashboard information modal, from the branch:
interlok-ui (feature/INTERLOK-3251_Dashboard_Management_Information)

screenshot:
screenshot-localhost_8080-2020 07 22-12_40_35

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/management/ManagementComponentInfo.java
+++ b/interlok-core/src/main/java/com/adaptris/core/management/ManagementComponentInfo.java
@@ -1,18 +1,21 @@
 package com.adaptris.core.management;
 
-import com.adaptris.core.ComponentState;
 import java.io.Serializable;
 
+import com.adaptris.core.ComponentState;
+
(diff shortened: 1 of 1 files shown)


quotidian-ennui:
@sebastien-belin-adp  @higgyfella  -> we should probably ask ourselves as to "why" Managementcomponent is even a member of that class ... since

* MgmtComponentInfo must be serializable so it can be transported via JMX
* ManagementComponent is never going to be serializable (that way lies madness)...

So we should probably get rid of it - I raised this as INTERLOK-3376

PR #496, opened 2020-08-04
feature/INTERLOK-3090

albinoloverats:
## Motivation

Use of generics allows for compile time checks that the correct objects are being encoded/decoded.

## Modification

Add generics to AdaptrisMessageEncoder and its implementations.

## Result

Checking that the encoder is a particular type and then casting is no longer necessary. Further thoughts and work may be desired for changes to AdaptrisMessageWorkerImp.


quotidian-ennui:
@albinoloverats I'm a bit confused as to why code-cov seems to think that the coverage delta on the MimeEncoders has gone down (e.g. MimeEncoder goes down by 5.99%?) -> is it worth checking this out, or is it just a function of timing and all that since the diff coverage is 100%.

PR #497, opened 2020-08-04
INTERLOK-3286 javax validation config checker

quotidian-ennui:
## Motivation

At the moment `java -jar lib/interlok-boot.jar -configcheck` doesn't actually validate that the configuration is correct; and it should.

## Modification

- Add a JavaxValidationChecker that uses Validation.buildDefaultValidatorFactory()
- Add a shared hierarchy for DeserializationCheck + JavaxValidationCheck so that we have common behaviour to get to the Adapter object.
   - The config checkers no longer delegate to the AdapterRegistry but instead manage how to get to the "configuration" themselves. It would have been better to delegate it to AdapterBuilder but that's a weird rats nest.
- Changed the tests so that we don't mock BootstrapProperties since there is odd behaviour from mockito around returning null after multiple mock invocations; I suspect that it's because of the class member mocking but life is too short to go unpicking mocks, and since bootstrapProperties doesn't need to be mocked; we can just have a real class that extends it.

## Result

If you have this as your adapter config 
```
<adapter>
  <shared-components>
    <connections>
    </connections>
    <services>
      <service-list>
        <unique-id>shared-service</unique-id>
        <services>
          <add-timestamp-metadata-service>
            <unique-id>timestamp with blank metadata key</unique-id>
            <metadata-key></metadata-key>
          </add-timestamp-metadata-service>
          <payload-hashing-service>
            <unique-id>payload-hash-with-no-algo</unique-id>
          </payload-hashing-service>
        </services>
      </service-list>
    </services>
  </shared-components>
</adapter>
```

then it should fail validation because payload-hashing-service requires an algorithm + metadata key. Similarly add-timestamp-metadata-service should have a metadata-key set but we have explicitly set it to be "blank".

Validation errors are always considered to be hard-failures (i.e. they're not warnings).

```
$ java -jar lib/interlok-boot.jar -configcheck
Bootstrap of Interlok 3.11-SNAPSHOT(2020-08-04:INTERLOK-3286-javax.validation-on-config-check) complete
TRACE [main] [c.a.c.m.BootstrapProperties] Properties resource is [bootstrap.properties]
.... some logging skipped

Configuration loading test:
Passed.

Classpath duplication check.:
Warnings found;
Possible duplicates found.  Re-run with '-Dinterlok.bootstrap.debug=true' for more details.

javax.validation Checks:
Failed with exceptions:
Interlok Validation Error: [sharedComponents.services[0].services[0].metadataKey]=[Must provide a metadata key]
Interlok Validation Error: [sharedComponents.services[0].services[1].metadataKey]=[must not be blank]
Interlok Validation Error: [sharedComponents.services[0].services[1].hashAlgorithm]=[must not be blank]

Shared connection check.:
Passed.

Config check only; terminating

chanl3@lhrm32445 ~/work/runtime/ant-nightly
$ echo $?
1

```

## Testing

see above.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/management/config/JavaxValidationChecker.java
+++ b/interlok-core/src/main/java/com/adaptris/core/management/config/JavaxValidationChecker.java
@@ -1,0 +1,46 @@
+package com.adaptris.core.management.config;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import javax.validation.ConstraintViolation;
+import javax.validation.Validation;
+import javax.validation.Validator;
+import javax.validation.ValidatorFactory;
+import com.adaptris.core.Adapter;
+import com.adaptris.core.CoreException;
+import lombok.NoArgsConstructor;
+
+@NoArgsConstructor
+public class JavaxValidationChecker extends AdapterConfigurationChecker {
+
+  private static final String FRIENDLY_NAME = "javax.validation Checks";
+  private static final ValidatorFactory JAVAX_VALIDATOR_FACTORY =
+      Validation.buildDefaultValidatorFactory();
+
+
+  @Override
+  protected void validate(Adapter adapter, ConfigurationCheckReport report) {
+    try {
+      Validator validator = JAVAX_VALIDATOR_FACTORY.getValidator();
+      // There are no warnings; everything is a hard failure.
+      report.setFailureExceptions((violationsToException(validator.validate(adapter))));
+    } catch (Exception ex) {
+      report.getFailureExceptions().add(ex);
+    }
+  }
+
+  @Override
+  public String getFriendlyName() {
+    return FRIENDLY_NAME;
+  }
+
+  private List<Exception> violationsToException(Set<ConstraintViolation<Adapter>> violations) {
+    List<Exception> result = new ArrayList<>();
+    for (ConstraintViolation v : violations) {
+      result.add(new CoreException(String.format("Interlok Validation Error: [%1$s]=[%2$s]",
+          v.getPropertyPath(), v.getMessage())));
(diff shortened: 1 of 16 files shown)


interlok-core/src/main/java/com/adaptris/core/management/config/JavaxValidationChecker.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  @Override
+  public String getFriendlyName() {
+    return FRIENDLY_NAME;
+  }
+
+  private List<Exception> violationsToException(Set<ConstraintViolation<Adapter>> violations) {
+    List<Exception> result = new ArrayList<>();
```
sebastien-belin-adp, line 39: Not a one liner?

```
  private List<Exception> violationsToException(Set<ConstraintViolation<Adapter>> violations) {
    return violations.stream().map(this::violationsCoreException).collect(Collectors.toList());
  }

  private CoreException violationsCoreException(ConstraintViolation<?> violation) {
    return new CoreException(
        String.format("Interlok Validation Error: [%1$s]=[%2$s]", violation.getPropertyPath(), violation.getMessage()));
  }
```

quotidian-ennui: You're right, but I had different issues with mocking ;) -> I shall change that to use a lambda (however, I haven't really loved the :: syntax so it might really be a one line (!).

quotidian-ennui: https://github.com/adaptris/interlok/pull/497/commits/f92fac6d171187cad1163e3c25952e08e28f8fa9

PR #513, opened 2020-08-26
Async producer now handles success and failure callbacks.

aaron-mcgrath-adp:
## Motivation

After recently adding onAdaptrisMessage(message, success, failure) we have needed to update the async producer to take advantage of these callbacks when a future event of success or failure comes in.

## Modification

The JmsAsyncProducer now uses a common method to that of SolaceJcsmpProducer in that it has a JMS 2.0 callback event handler (CompletionListener) registered.  The producer will register the JMS message with the adaptris message callbacks in the event handler.  Later when a success/failure event is caught we recall the adaptris message with the callbacks in it and execute them.

## Result

The async producer is now fully functional according to the JMS 2.0 spec.  In testing it's just as fast producing to WMQ (which supports jms 2.0 from version 8.0) as the solace native stuff.

## Testing

You'll need a JMS 2.0 compliant broker like Artemis or WMQ.
Simply configure this producer instead of a JMS 1.1 producer and away you go!
You'll see additional trace logging informing you messages have been added to the pile of un ack'd messages, then further logging telling you of the success/failure events received.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/jms/JmsAsyncProducer.java
+++ b/interlok-core/src/main/java/com/adaptris/core/jms/JmsAsyncProducer.java
@@ -1,19 +1,19 @@
 package com.adaptris.core.jms;
 
-import javax.jms.CompletionListener;
 import javax.jms.JMSException;
 import javax.jms.Message;
-import javax.validation.Valid;
-import javax.validation.constraints.NotNull;
+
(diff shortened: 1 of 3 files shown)


quotidian-ennui (approved):
I got a bit confused reading the code initially; cos I thought the eventHandler was in fact the EventHandler!.

PR #514, opened 2020-08-26
INTERLOK-2611 split many messages

quotidian-ennui:
## Motivation

If have a message that has many many messages when split (e.g. a LineCountSplitter on a 80k line file) then you end up with a list of 80k futures. This is undesirable from an object count of point of view vis-a-vis object creation, since this effectively doubles the number of object references (1 for the split message, and 1 for the future).

## Modification

- Add a Consumer<Exception> callback to MessageSplitterServiceImp that concrete sub-classes know about
- Make the concrete subclasses use the callback on success or failure which is then tracked to make sure that we throw
an exception as required.
- Add a "wait-while-busy (default=false)" mode to pooling-message-splitter so that it proactively checks the size of the object pool before submitting; waits a while if num-active >= max-allowed.
    - This has the effect of ensuring that 'max-threads' number of AdaptrisMessages are in memory, since the splitter itself will be a lazy iterable in most cases (line-count-splitter / large-json-array-splitter) etc.
- Add explicit tests for ServiceExceptionHandler since we shouldn't rely on the kindness of strangers for unit-testing it.

Ideally we would be able to use the same methodology for reducing the object count when using PoolingSplitJoin or similar (but from a cursory glance we can't since we need to keep a reference to all the Futures so that we can cancel them if the timeout expires -> there's no such requirement for a split).

## Result

No configuration changes; but if people are using PoolingMessageSplitterService in a constrained environment, they might see some improvements vis a vis memory use.

## Testing

Only needed to change 1 test (BasicMessageSplitter) to remove the custom Future that we had (the test was only for coverage); so we should be able to trust the tests...
Probably want to test it with a massive file to check "memory use" before and after .

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/splitter/AdvancedMessageSplitterService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/splitter/AdvancedMessageSplitterService.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2015 Adaptris Ltd.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


sebastien-belin-adp (approved):
I didn't do any real performance check or comparison but it works fine with lots of splits.

quotidian-ennui:
Some numbers (but no pretty pictures; they're in the ticket proper)

With this PR for a large-json-splitter that splits to give ~52k json objects, written out to disk.

No-Config change : [7594] ms, memory use ~ 400/500Mb, 2nd run, no sig change
Wait-While-Busy :  [324323] ms, memory use ~ 100/150Mb (but note the 50x performance hit), didn't do a 2nd run cos I'm not waiting > 5minutes
Pre-PR : 758]ms : memory use ~600Mb , depending on GC; next run, memory use went up to 1.1Gb (futures not yet eligible for GC?)

PR #515, opened 2020-08-27
INTERLOK-3425 pooling splitter wait while busy

quotidian-ennui:
## Motivation

https://github.com/adaptris/interlok/pull/514 caused a drop in performance while wait-while-busy=true. For a simplistic test, generating 52k split messages, we ended up taking 50times longer than wait-while-busy=false.  Which is expected if we just do the maths.

- There are 52k messages coming out of the split.
- Each split message "takes no time to process"; since it's just a raw write to disk, so we can expect the entire thread pool to be emptied at the same time (near as dammit).
- This means that the total wait time is possibly (52k / 10 (the size of the pool)) * 100 -> which is 520k ms or ~10 minutes.
- The performance measuring shows that the time to process is 320k ms, so we're somewhat faster, however the bulk of this time is spent in Thread.sleep() for no reason.

Needed to fix that, otherwise there's no point to that flag at all.

## Modification

Use wait()/notifyAll() rather than naively sleeping for 100ms when waiting for an object to be returned back to the object pool.

## Result

if wait-while-busy=true, then performance should be comparable to wait-while-busy=false. GC happens when it wants to happen, so memory use characteristics appear to be better (not precisely measured).

## Testing

The existing tests still work, so I anticipate no regressions. Configure a massive file to split (e.g. a line count splitter + file of 80k lines), and the performance with 'wait-while-busy=true' should be comparable to 'wait-while-busy=false' but memory use will go up & down nicely if you hard-fix the memory with `-Xmx`

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/splitter/PoolingMessageSplitterService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/splitter/PoolingMessageSplitterService.java
@@ -17,7 +17,6 @@ package com.adaptris.core.services.splitter;
 
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Future;
 import java.util.function.Consumer;
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.pool2.impl.GenericObjectPool;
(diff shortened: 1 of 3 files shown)


quotidian-ennui:
Well, thoroughly strange; it's 100% coverage on pooling splitter on my mac via jacoco

sebastien-belin-adp (approved):
I ran the poling splitter on a 80k+ lines file with waitWhileBusy set to false and then set to true and monitored it with jconsole.
The memory, cpu usage and speed are very similar.

PR #520, opened 2020-09-01
Feature/interlok 3420 jms consumer synch mode

sebastien-belin-adp:
## Motivation

Now that we can use JMS 2.0 async producers and native API's that also have multi threaded future callback mechanisms we need to have a synchronous consumer that will allow separate threads to acknowledge messages.

## Modification

I abstracted some code from JmsPollingConsumerImpl to a new BaseJmsPollingConsumerImpl (I ran out of idea for the name).
Created a new JmsSyncConsumer class that extend BaseJmsPollingConsumerImpl and it's basically a polling consumer that use an adapter JmsConnection instead of initiating/closing a jms connection everytime.

I also added a new JmsMessageConsumerFactory class the create a JMS MessageConsumer depending on destination type, subscription id, shared consumer id ...
This new class is used in JmsConsumer and in the new JmsSyncConsumer.

Added some test for all this new code.

## Result

We now have a new synch jms consumer that doesn't reset the connection and that can be used with the new JmsAsyncProducer.

## Testing

All the existing JMS consumer tests should still pass and the test in JmsSyncConsumerTest and JmsSyncConsumerTest should also pass.
It should look good in the UI and work fine in an adapter:

```
<consumer class="jms-sync-consumer">
  <unique-id>MyJmsSyncConsumer</unique-id>
  <poller class="fixed-interval-poller">
    <poll-interval>
      <unit>SECONDS</unit>
      <interval>5</interval>
    </poll-interval>
  </poller>
  <acknowledge-mode>CLIENT_ACKNOWLEDGE</acknowledge-mode>
  <message-translator class="auto-convert-message-translator">
    <jms-output-type>Text</jms-output-type>
  </message-translator>
  <endpoint>jms:topic:myTopic</endpoint>
</consumer>
```

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/jms/BaseJmsPollingConsumerImpl.java
+++ b/interlok-core/src/main/java/com/adaptris/core/jms/BaseJmsPollingConsumerImpl.java
@@ -1,0 +1,281 @@
+package com.adaptris.core.jms;
+
+import static com.adaptris.core.jms.NullCorrelationIdSource.defaultIfNull;
+
+import java.util.concurrent.TimeUnit;
+
+import javax.jms.IllegalStateException;
+import javax.jms.JMSException;
+import javax.jms.Message;
+import javax.jms.MessageConsumer;
+import javax.jms.Session;
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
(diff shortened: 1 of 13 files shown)


aaron-mcgrath-adp (approved):
Nice!
Works great.

PR #524, opened 2020-09-08
INTERLOK-3424 pooling split join performance characteristics

quotidian-ennui:
## Motivation

When you have a message-splliter implementation that returns you a huge number of split messages (e.g. in the order of 80k+) then you end up in a situation where you have a List<> that contains 80k entries. This is generally "handled" but probably isn't efficient on constrained environments. 

The code is effectively (there is nuance but not much) which effectively implies that you have at least the total split count worth of objects sitting in eden space.

```java
Iterable<AdaptrisMessage> list = splitter.splitMessages(); // might be a list, might be lazy iterating
List<Future> futures = doSplit(list); // definitely a list of 80k messages now
List<AdaptrisMessage> done = futures.get();
aggregator.joinMessage(original, done);
```

## Modification

- MessageAggregator interface adds a `default void aggregate(AdaptrisMessage original, Iterable<AdaptrisMessage> msgs)` method which just listifies and calls joinMessage, so by default, no aggregators need to change, but won't give you any performance benefit
- Prototype `PooledSplitJoinService` which behaves slightly different to PoolingSplitJoin. 
  - poolsize is "fixed", and warmed up on start.
  - object.wait()/notify() to inform us when the pool is available
  - Messages are submitted to a pool in a separate thread, which means that aggregating starts happening at the same time as the split.
  - Timeout is handled manually rather than via Executors.submit()
- Additional `ServiceErrorHandler` implementations which allow us to replicate the behaviour of `PoolingFutureExceptionStrategy`
  - If you're going to use a non-default strategy, since aggregation might happen as soon as messages are available; you'll have to be careful about "filtering" conditions.

## Result

- There is a PooledSplitJoinServiceTest which just uses AppendingMessageAggregator to test functionality. Tests pass, which means the interface change is OK.
- PooledSplitJoinService is available to configure, along with supporting `ServiceErrorHandler` implementations which replicate the deprecated `PoolingFutureExceptionStrategy` implementations.
- Since configuration is different, we deprecate SplitJoinService & PoolingSplitJoinService with a view for removing it in 4.0.0
- Usage of Deprecated classes will result in a warning during "prepare()" from 3.11.1

## Testing

Select the PooledSplitJoinService, and monitor the behaviour (via jconsole?) on very very large split messages.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/splitter/PooledSplitJoinService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/splitter/PooledSplitJoinService.java
@@ -63,20 +63,43 @@ import lombok.Setter;
 /**
  * Implementation of the Splitter and Aggregator enterprise integration pattern.
  *
+ * <p>
+ * This supersedes {@link SplitJoinService} and {@link PoolingSplitJoinService} as our preferred
+ * service for handling split/aggregation since it has more predictable peformance characteristics
+ * in constrained environments where the numbers of messages that are generated and aggregated can
+ * be large/unknown.
+ * </p>
+ * <p>
+ * This service splits a message according to the configured {@link MessageSplitter} implementation,
+ * executes the configured {@link Service} and subsequently aggregates all the messages back using
+ * the configured {@link MessageAggregator} implementation
+ * <p>
+ * <p>
+ * A pool of {@link com.adaptris.core.Service} instances is maintained and re-used for each message;
+ * the cost of initialisation for the wrapped service, is incurred during this service's
+ * initialisation phase.
+ * </p>
+ * <p>
+ * Aggregation may start happening as soon as messages are available to be aggregated (i.e. a split
+ * message has been operated on) using the
+ * {@link MessageAggregator#aggregate(AdaptrisMessage, Iterable)} method. Performance characterstics
+ * will largely depend on how the splitter and aggregator implementations iterate over the messages.
+ * </p>
  *
- * @config fixed-split-join-service
+ * @config pooled-split-join-service
  */
-@XStreamAlias("fixed-split-join-service")
+@XStreamAlias("pooled-split-join-service")
 @AdapterComponent
 @ComponentProfile(
     summary = "Split a message and then execute the associated services on the split items, aggregating the split messages afterwards",
     since = "3.11.1", tag = "service,splitjoin")
 @DisplayOrder(order =
 {
-    "splitter", "service", "aggregator", "timeout", "poolsize"
+    "splitter", "service", "aggregator", "timeout", "poolsize", "sendEvents",
+    "serviceErrorHandler"
 })
 @NoArgsConstructor
-public class FixedSplitJoinService extends ServiceImp implements EventHandlerAware, ServiceWrapper {
+public class PooledSplitJoinService extends ServiceImp implements EventHandlerAware, ServiceWrapper {
 
   private static final long DEFAULT_TTL = TimeUnit.MINUTES.toMillis(10L);
   private static final int DEFAULT_POOLSIZE = 10;
@@ -128,10 +151,10 @@ public class FixedSplitJoinService extends ServiceImp implements EventHandlerAwa
   @InputFieldDefault(value = "false")
   @Getter
   @Setter
+  @AdvancedConfig
   private Boolean sendEvents;
   @Getter
   @Setter
-  @AdvancedConfig
   @InputFieldDefault(value = "10")
   private Integer poolsize;
 
--- a/interlok-core/src/main/java/com/adaptris/core/services/splitter/PooledSplitJoinService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/splitter/PooledSplitJoinService.java
@@ -82,8 +82,9 @@ import lombok.Setter;
  * <p>
  * Aggregation may start happening as soon as messages are available to be aggregated (i.e. a split
  * message has been operated on) using the
- * {@link MessageAggregator#aggregate(AdaptrisMessage, Iterable)} method. Performance characterstics
- * will largely depend on how the splitter and aggregator implementations iterate over the messages.
+ * {@link MessageAggregator#aggregate(AdaptrisMessage, Iterable)} method. Performance
+ * characteristics will largely depend on how the splitter and aggregator implementations iterate
+ * over the messages.
  * </p>
  *
  * @config pooled-split-join-service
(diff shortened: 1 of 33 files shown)


quotidian-ennui:
Some raw numbers, not scientific... Memory peak seems lower, obviously GC can happen on an as-is basis.

```
Fixed Split Join

DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [3777] ms
DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [3184] ms
DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [3033] ms

Memory Peak : 888Mb

# Pooling Split Join

DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [6478] ms
DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [3934] ms
DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [4177] ms
DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [4266] ms

Memory Peak : 1210Mb
```

quotidian-ennui:
If we constrain the environment to use -Xmx256m then we get these perf numbers (which are broadly comparable) then FixedSplitJoin gives us

```
DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [3790] ms
DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [3388] ms
DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [3343] ms
DEBUG [json-array-split-join@filesystem] [c.a.c.RandomIntervalPoller] time to process [1] messages [3443] ms

Memory Peak : ~180Mb
```

If we attempt to do the same with PoolingSplitJoin, then it hammers the CPU for a while, and eventually fails with a OOM error.

```
DEBUG [json-array-split-join@filesystem] [c.a.c.ServiceList] Executing doService on [PoolingSplitJoinService(split-join)]
TRACE [json-array-split-join@filesystem] [c.a.c.s.s.j.LargeJsonArraySplitter] Using message factory: class com.adaptris.core.DefaultMessageFactory
Exception in thread "Scanner-0" java.lang.OutOfMemoryError: Java heap space
	at java.util.RegularEnumSet.iterator(RegularEnumSet.java:75)
	at java.nio.file.FileTreeWalker.<init>(FileTreeWalker.java:179)
	at java.nio.file.Files.walkFileTree(Files.java:2661)
	at org.eclipse.jetty.util.Scanner.scanFiles(Scanner.java:670)
	at org.eclipse.jetty.util.Scanner.scan(Scanner.java:640)
	at org.eclipse.jetty.util.Scanner$1.run(Scanner.java:558)
	at java.util.TimerThread.mainLoop(Timer.java:555)
	at java.util.TimerThread.run(Timer.java:505)
Exception in thread "qtp650439389-40" Exception in thread "Job_Executor0" java.lang.OutOfMemoryError: Java heap space
java.lang.OutOfMemoryError: Java heap space
Exception in thread "Job_Executor4" java.lang.OutOfMemoryError: Java heap space
^C^COpenJDK 64-Bit Server VM warning: Exception java.lang.OutOfMemoryError occurred dispatching signal SIGINT to handler- the VM may need to be forcibly terminated
Exception in thread "Job_Executor5" java.lang.OutOfMemoryError: Java heap space
^CRunning ShutdownHandler, please wait...
INFO  [Shutdown Handler] [j.l.Thread] Running ShutdownHandler...
```

quotidian-ennui:
At the moment, there's a counting-exception-handler that has the expect split count set once all the messages have been submitted. It's not the most elegant solution (but it does work).

We won't get CME because the iterator doesn't remove, or do anything funky; equally it's written so that it fails if you try to get the iterator more than once.

```
Iterable<Msg> iterable = splitter().split(original);
Iterator i = iterable.iterator()
Iterator j = Iiterable.iterator()  // This will fail with a runtime exception.
```

It's arguable that the iterator itself is not thread save in the sense that hasNext()/next() probably aren't thread safe, but I don't think we need to be in this specific instance.

sebastien-belin-adp (commented):
I've compared the old PoolingSplitJoin the new PoolingSplitJoin and the PooledSplitJoin.
With a 80k+ json file the differences are not massive but it seems that globally the PooledSplitJoin is better than the Old PoolingSplitJoin and the New PoolingSplitJoin. However the Old PoolingSplitJoin seems to perform better than the new one.
The PooledSplitJoin tends to be faster (about 5s) than the New PoolingSplitJoin and faster (about 2s) than the  Old PoolingSplitJoin.
The PooledSplitJoin use less heap memory than the Old PoolingSplitJoin which use les than the New PoolingSplitJoin. But the Old PoolingSplitJoin heap memory stay full for longer.
In term of CPU usage it seems to depends (probably my computer). But the Old PoolingSplitJoin in average uses less CPU.

But looking at the changes the PooledSplitJoin advantage will probably increase with the file size.

Other than that the PooledSplitJoin makes sense and looks good in the UI.

And there is a warning in PooledSplitJoinService: The variable timeoutExpired in PooledSplitJoinService#constructAdaptrisMessage is not used anywhere.

interlok-core/src/main/java/com/adaptris/core/services/splitter/PooledSplitJoinService.java
```diff
@@ -0,0 +1,437 @@
...
+ * <p>
+ * <p>
+ * A pool of {@link com.adaptris.core.Service} instances is maintained and re-used for each message;
+ * the cost of initialisation for the wrapped service, is incurred during this service's
+ * initialisation phase.
+ * </p>
+ * <p>
+ * Aggregation may start happening as soon as messages are available to be aggregated (i.e. a split
+ * message has been operated on) using the
+ * {@link MessageAggregator#aggregate(AdaptrisMessage, Iterable)} method. Performance characterstics
```
sebastien-belin-adp, line 85: Javadoc typo: It should be "Performance characteristics" instead of "Performance characterstics"

interlok-core/src/main/java/com/adaptris/core/services/splitter/PooledSplitJoinService.java
```diff
@@ -0,0 +1,437 @@
...
+   */
+  @Valid
+  @Getter
+  @Setter
+  private TimeInterval timeout;
+  @InputFieldDefault(value = "false")
+  @Getter
+  @Setter
+  @AdvancedConfig
+  private Boolean sendEvents;
```
sebastien-belin-adp, line 155: Javadoc is missing. Maybe we can use the same as in AdvancedMessageSplitterService#setSendEvents

sebastien-belin-adp (approved):
Javadoc fixed :)

PR #529, opened 2020-09-14
INTERLOK-3437 Shared component checker now uses pre-processors

quotidian-ennui:
## Motivation

If your configuration is 
```
<adapter xmlns:xi="http://www.w3.org/2001/XInclude">
  <shared-components>
     <xi:include href="connections.xml"/>
  </shared-components>
  </channel-list>
    <channel>
          <lookup-name>something defined in connections.xml</lookup-name>
       </consume-connection>
    </channel>
  </channel-list>
</adapter>
```
Because you love xincludes then java -jar lib/interlok-boot.jar does not work since the shared confugiration checker doesn't believe in pre-processors.

## Modification

Add a helper method into ConfigPreProcessorLoader that does the right thing as a helper method which basically looks like this : 
```
  static String loadInterlokConfig(BootstrapProperties config) throws Exception {
    String result = "";
    try (InputStream in = config.getConfigurationStream()) {
      result = IOUtils.toString(in, Charset.defaultCharset());
    }
    return new DefaultPreProcessorLoader().load(config).process(result);
  }
```

- Make the config checkers use it.

## Result

No change for most people; but xinclude lovers can now use the parent gradle.

## Testing

See motivation above

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/config/ConfigPreProcessorLoader.java
+++ b/interlok-core/src/main/java/com/adaptris/core/config/ConfigPreProcessorLoader.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2015 Adaptris Ltd.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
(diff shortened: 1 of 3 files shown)


mcwarman (approved):
Tested and working as expected

PR #534, opened 2020-09-30
Interlok 3455

aaron-mcgrath-adp:
## Motivation

Creating a durable subscriber was broken.  However, we were probably getting away with it while testing against ActiveMq because it didn't check to make sure the subscription name was null/empty.  The artemis broker does however which threw an pointed out this error.

## Modification

We're now using the  subscription name when creating a durable subscriber in the JmsMessageConsumerFactory.

I've also used a static embedded broker and set the connection reconnects in an attempt to fix timing issues in the JmsConsumerTest.

## Result

Fingers crossed the unit tests won't fail periodically.

Durable subscribers can now be created correctly.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/jms/JmsMessageConsumerFactory.java
+++ b/interlok-core/src/main/java/com/adaptris/core/jms/JmsMessageConsumerFactory.java
@@ -50,13 +50,13 @@ public class JmsMessageConsumerFactory {
           } else {
             log.trace("Creating new durable consumer.");
             consumer = ((ConsumerCreator) (session, dest, filterExpression) -> session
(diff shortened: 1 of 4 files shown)


quotidian-ennui (approved):

quotidian-ennui:
I realise there's a side effect to this change; we're now not using activemq at all in JmsConsumerTest; we've basically switched to Artemis for some JMS 2.0 action + JndiImplementation. We're using EmbeddedActiveMQ elsewhere so this change has no consequence other than giving us (positive) potential coverage of side-effects of using the Artemis client jars.

Artemis is meant to have protocol compatibility with ActiveMQ so I wonder if it's worth explicitly testing if `BasicActiveMQImpl/activemq-client.jar/tcp://localhost:61616` will work with an Artemis broker (I suspect it will, but only in JMS 1.1 mode).

PR #540, opened 2020-10-05
INTERLOK-3461 Add support for the no-args constructor in FileFilterBuilder

quotidian-ennui:
## Motivation

You may have a custom filter that doesn't "require" a filter-expression. This is currently not possible with the behaviour of FileFilterBuilder since w/o an filter expression you always have an "accept-all" filter.

The underlying behaviour here was reported against a FtpConsumer (SFTP) that tried to process sub-directories.
- They tried to use FileFileFilter but that was ignored because an empty-filter-expression.
- They could have used a regex negated look ahead along the lines of `^(?!dir_to_ignore).*$` but this seems somewhat obtuse.

## Modification

- RemoteBlobFilterWrapper now defaults to RegexFileFilter to avoid issues during tests.
- FileFilterBuilder now attempts to create a FileFilter if filter-expr is blank, provided the filter-imp != RegexFileFilter
- Documented the behaviours in the javadocs.
- Removed / deprecate constants in FsConsumer/FtpConsumer
- Add IsFileFilter as an explicit filter impl.
- Modify TempFileUtils to support auto-populating the file with data.

## Result

- You can now configure a `com.adaptris.core.fs.IsFileFilter` as your filter implementation in Ftp / Fs Consumers.
- You cannot configure `org.apache.commons.io.filefilter.FileFileFilter` as your filter implementation (that hasn't changed from existing behaviour, since FileFileFilter doesn't have a public no-args constructor.

## Testing

- Create a "work directory" for consuming from (for fs).
- Create a subdirectory of that work directory (_work-should-be-ignored)
- Configure your fs-consumer with com.adaptris.core.fs.IsFileFilter as your filter-implementation
- apply & wait : it should always give you 0 files processed.
- create some files in the normal work directory...
- They will be processed but the sub-dir will be ignored.

### Backwards compat testing.
- Configuring an org.apache.commons.io.filefilter.FileFileFilter will fail during prepare/init when you apply it since it is missing a public no-args constructor

```
Caused by: java.lang.RuntimeException: java.lang.IllegalAccessException: Class com.adaptris.interlok.util.FileFilterBuilder can not access a member of class org.apache.commons.io.filefilter.FileFileFilter with modifiers "protected"
	com.adaptris.interlok.util.FileFilterBuilder.build(FileFilterBuilder.java:69)
	com.adaptris.core.fs.FsHelper.createFilter(FsHelper.java:154)
	com.adaptris.core.fs.FsConsumerImpl.init(FsConsumerImpl.java:314)
```


sebastien-belin-adp (approved):
Works fine, tests pass and changes make sense.

Only one thing (but not exactly related) in the UI I think **File Filter Implementation** and **Filter Expression** should be next to each other.
That probably mean modifying @DisplayOrder for FsConsumer, LargeFsConsumer, TraversingFsConsumer, NonDeletingFsConsumer and MovingNonDeletingFsConsumer.

PR #541, opened 2020-10-06
INTERLOK-3460 xpath service attribute node support

Co-Din:
## Motivation

Why am I doing this

I discovered that when retrieving a single attribute node in the XPath Service the service would simply throw an exception rather than return the value

## Modification

What did I change

I updated the Xpath service to resolve attribute nodes and updated the service's tests to reflect this.

## Result

The user can now resolve single attribute nodes with the Xpath service

## Testing

The tests have been updated and cover the changes.
You can also build the project, drop the jar into a working adapter and query an xpath attribute node to confirm

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/path/XPathService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/path/XPathService.java
@@ -12,7 +12,7 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
-*/
+ */
 
 package com.adaptris.core.services.path;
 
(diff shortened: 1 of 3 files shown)


interlok-core/src/main/java/com/adaptris/core/services/path/XPathService.java
```diff
@@ -248,11 +248,17 @@ private String serializeNode(NodeList nodeList) throws TransformerException {
...
+	    StringWriter stringWriter = new StringWriter();
+		StreamResult xmlOutput = new StreamResult(stringWriter);
+	    Transformer transformer = TransformerFactory.newInstance().newTransformer();
+	    transformer.setOutputProperty(OutputKeys.OMIT_XML_DECLARATION, "yes");
+	    if (source.getNode().getNodeType() != Node.ATTRIBUTE_NODE) {
+	        transformer.transform(source, xmlOutput);
+	      } else {
+	        stringWriter.write(source.getNode().getNodeValue());
+	      }
+	    return stringWriter.toString();
```
quotidian-ennui, line 261: Spaces not tabs... (when I view the while file, this sticks out like a sore thumb).

quotidian-ennui:
What's the "point" of the change; I mean I understand the point of the change, but why has it manifested itself?

- Do we get the problem with XPathMetadataService if we run the xpath + xml against that?
- Is it just XpathService because it does selectNodeList and then tries to render it that as text?

- Could you write a test for XpathMetadataService that does the same test as this (that's the one that most people will be using).

interlok-core/src/main/java/com/adaptris/core/services/path/XPathService.java
Co-Din: Apologies, just updated the indentations to two spaces

Co-Din:
@quotidian-ennui 

Manifested while trying to create an example using the Xpath service for documentation, the service couldn't resolve attribute nodes.

We don't get the same problem with XpathMetadata, in fact we can find and will resolve all values matching the node attribute.

 Is it just XpathService because it does selectNodeList and then tries to render it that as text?
It seems when I followed in debug that in XpathMetadata when it's configuring the Xpath Metadata query it's actually setting it to a MultiItemConfiguredXpathQuery rather than ConfiguredXpathQuery though I'm haven't managed to discern where this is happening

Could you write a test for XpathMetadataService that does the same test as this (that's the one that most people will be using).

Now added two tests for single and multiple matches to attribute node queries

quotidian-ennui (approved):

PR #544, opened 2020-10-08
ADR-0009 proxy message factory.

quotidian-ennui:
## Motivation

There is a specific use-case where Interlok may have possibly been the wrong choice. It's essentially behaving as a passthrough stream of data into Elasticsearch. It's arguable that this is the wrong approach but Interlok Hammer

What can we do to mitigate the IOPS cost that happens when we download an S3 blob and then essentially call some other relatively expensive operation with the contents of the blob.

Pre Merge : https://github.com/adaptris/interlok/blob/ADR-0009-mutable-messages/docs/adr/0009-mutable-messages.md
Post Merge : https://github.com/adaptris/interlok/blob/develop/docs/adr/0009-mutable-messages.md

### DIFF
--- a/docs/adr/0009-mutable-messages.md
+++ b/docs/adr/0009-mutable-messages.md
@@ -28,8 +28,7 @@ What are the options that we can provide to mitigate the problem.
 
 ## Decision Outcome
 
-Chosen option: "[option 1]", because [justification. e.g., only option, which meets k.o. criterion decision driver , which resolves force force , ... , comes out best (see below)].
-
+...
 
 ### Do Nothing
 
--- a/docs/adr/0009-mutable-messages.md
+++ b/docs/adr/0009-mutable-messages.md
@@ -25,6 +25,7 @@ What are the options that we can provide to mitigate the problem.
 - Do Nothing
 - Custom JSON parsing message factory
 - Mutable Message
+- Streaming Workflow
 
 ## Decision Outcome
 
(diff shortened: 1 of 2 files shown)


quotidian-ennui:
The more I think about this, the more I think that while a mutable message factory might be useful generically, it may not be useful at all in this particular use-case since effectively there's 4 IOPS costs

- Download the Blob -> network IOPS
- Write the blob to disk (since it is large) -> disk IOPS
- Read the blob from disk -> disk IOPS
- batched index to elastic search -> network IOPS

By having a s3-readonly-factory we only semi-avoid the disk IOPS which may not be the "high cost", unless we have repeated operations.

Equally if we look to the javadocs for S3 SDK, then we will probably ultimately end up calling `AmazonS3Client#getObject()` since TransferManager only has download() to a file. This then leads us down to S3Object#getObjectContent has javadoc notes : 

```
    /**
     * Gets the input stream containing the contents of this object.
     *
     * <p>
     * <b>Note</b>: The method is a simple getter and does not actually create a
     * stream. If you retrieve an S3Object, you should close this input stream
     * as soon as possible, because the object contents aren't buffered in
     * memory and stream directly from Amazon S3. Further, failure to close this
     * stream can cause the request pool to become blocked.
     * </p>
     *
     * @return An input stream containing the contents of this object.
     *
     * @see S3Object#getObjectMetadata()
     * @see S3Object#setObjectContent(InputStream)
     */
```

mcwarman (approved):

quotidian-ennui:
No traction and no interest

PR #555, opened 2020-10-28
INTERLOK-3472 Stop processing Conditional loops

aaron-mcgrath-adp:
## Motivation

A recent QandA message suggested that our conditional loops (while and do-while) continue looping after a stop-processing-service.
ServiceList's do actually test for the metadata set by the stop-processing-service, so you could argue that even though the conditional loops will continue spinning, no services will be applied.
But what if you use a service that has nested services and one of those nested services is a stop-processing-service.

This behaviour doesn't make sense for the for-loop, considering that loop uses a different payload for each iteration.  Also each payload has the services run in parallel threads making the behaviour of stopping exactly as soon as one of the threads runs a stop-processing thread difficult to predict.  

I think this PR can be argued either way.

## Modification

CoreConstants now has a static function that is reused in place of having to know the core constants set by the stop-processing-service.  Each conditional loop executes this function to break free from the loop.  

We might even consider adding a new class CoreFunctions; with a static list of commonly used functions such as checking if there is exception data held within a message.

## Result

Conditional loops will break from their loop if the stop-processing-service has been run.

## Testing

Difficult to test considering eventual functionality doesn't really change.
New unit tests added though show the breaking of the loops.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/CoreConstants.java
+++ b/interlok-core/src/main/java/com/adaptris/core/CoreConstants.java
@@ -18,6 +18,8 @@ package com.adaptris.core;
 
 import java.util.function.Function;
 
+import org.apache.commons.lang3.BooleanUtils;
+
 import com.adaptris.validation.constraints.ConfigDeprecated;
 
 /**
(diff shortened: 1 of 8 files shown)


interlok-core/src/main/java/com/adaptris/core/CoreConstants.java
```diff
@@ -39,6 +41,23 @@
...
+   * @see #KEY_WORKFLOW_SKIP_PRODUCER
+   */
+  public static final String STOP_PROCESSING_VALUE = "true";
+  
+  /**
+   * A simply function check to test if the processing of the given message should stop.
+   */
+  public static final Function<AdaptrisMessage, Boolean> shouldStopProcessing = adaptrisMessage -> {
+    return STOP_PROCESSING_VALUE.equals(adaptrisMessage.getMetadataValue(STOP_PROCESSING_KEY));
+  };
```
quotidian-ennui, line 60: While I do like a `{}` block; is it not possible to simplify this to just
```
public static final Function<AdaptrisMessage, Boolean> shouldStopProcessing = 
  am -> BooleanUtils.toBoolean(am.getMetadataValue(STOP_PROCESSING_KEY))
```

Then we can just deprecate the STOP_PROCESSING_VALUE and say it can be any of `y|yes|on|true` since BooleanUtils does the case magic for us as well...

quotidian-ennui (approved):

PR #556, opened 2020-10-29
INTERLOK-3465 Trigger retries from jetty.

quotidian-ennui:
## Motivation

This is the changes for : https://github.com/adaptris/interlok/blob/develop/docs/adr/0008-restful-failed-message-retrier.md

## Modification

- Changed FailedMessageRetrier interface so that it no longer extends AdaptrisMessageListener; since there is only a single implementation `DefaultFailedMessageRetrier` across the entire codebase (including optionals) this should be a low-impact change
    - Might have a knock-on effect on the UI, since there is stuff in there explicitly about FailedMessageRetrier
- Added a RetryStore interface with a single public implementation `FilesystemRetryStore` which supports storing messages on the filesystem.
- Added a RetryFromJetty implementation
   - Reports on failed messages via a configurable endpoint (_/api/failed/list_). The default renderer is just a newline separated list of filenames.
   - Supports retrying of failed messages via a configurable endpoint (_/api/retry/'msgId'_)
   - Jetty connection is configurable since it's a normal 'Connection' but is defaulted to be `EmbeddedConnection`
- Add various `RetryStore*` services for completeness.
- __Fixed a long standing "bug" whereby Adapter never called prepare() on its FailedMessageRetrier__
  - By coincidence since everyone only ever uses DefaultFailedMessageRetrier; against the filesystem. It wasn't apparent because FS stuff doesn't need _prepare()_ and it was handling stuff behind the scenes.

The interface definition/behaviour has deviated somewhat from the original ADR which are listed below
- RetryStore defines a `getMetadata(String)` method; which allows us to not read the payload (since this might be large)
- RetryStore defines a `buildForRetry` method which takes a AdaptrisMessageFactory parameter; to avoid yet more AMF configuration.
- RetryFromJetty uses getMetadata() to shortcut checking whether the message has a workflow associated with it.
- RetryFromJetty figures out the workflow, and then gets the MessageFactory associated with that consumer, and uses that to subsequently invoke `buildForRetry()` with metadata + message factory since this means we don't need to care about whether the message is large or not.

## Result

- The com.adaptris.core.http.jetty.retry has 100% coverage for tests.
- RetryFromJetty is now available to configure as the FailedMessageRetrier.
- RetryFromJetty should be used in a conjunction with a `standard-processing-exception-handler` that uses `retry-store-write-service` as part of its exception chain. 

## Testing

How can I test this if I'm reviewing this -> This minimal configuration should allow for testing... (requires interlok-json because...)

```xml
<adapter>
  <shared-components>
    <connections>
      <jetty-embedded-connection>
        <unique-id>jetty</unique-id>
      </jetty-embedded-connection>
    </connections>
    <services/>
  </shared-components>
  <message-error-handler class="standard-processing-exception-handler">
    <unique-id>error-handler</unique-id>
    <processing-exception-service class="service-list">
      <unique-id>store-for-retry</unique-id>
      <services>
        <retry-store-write-message>
          <unique-id>store-for-retry</unique-id>
          <retry-store class="retry-store-filesystem">
            <base-url>file:///./fs/retries</base-url>
          </retry-store>
        </retry-store-write-message>
        <exception-report-service>
          <unique-id>generate-json-exception</unique-id>
          <exception-serializer class="exception-as-json"/>
        </exception-report-service>
        <jetty-response-service>
          <unique-id>send-500</unique-id>
          <http-status>500</http-status>
          <content-type>application/json</content-type>
        </jetty-response-service>
      </services>
    </processing-exception-service>
  </message-error-handler>
  <failed-message-retrier class="retry-via-jetty">
    <unique-id>silly-clarke</unique-id>
      <lookup-name>jetty</lookup-name>
    </connection>
    <report-builder>
      <report-renderer class="remote-blob-list-as-json"/>
      <content-type>application/json</content-type>
    </report-builder>
    <retry-store class="retry-store-filesystem">
      <base-url>file:///./fs/retries</base-url>
    </retry-store>
  </failed-message-retrier>
  <channel-list>
    <channel>
        <lookup-name>jetty</lookup-name>
      </consume-connection>
      <workflow-list>
        <standard-workflow>
          <consumer class="jetty-message-consumer">
            <unique-id>boring-brattain</unique-id>
            <path>/api/always-fail</path>
          </consumer>
            <unique-id>pedantic-montalcini</unique-id>
            <services>
              <throw-exception-service>
                <unique-id>always-fail</unique-id>
                <exception-generator class="configured-exception">
                  <message>Always Fails.</message>
                </exception-generator>
              </throw-exception-service>
            </services>
          </service-collection>
          <unique-id>always-fail</unique-id>
        </standard-workflow>
      </workflow-list>
      <unique-id>default-channel</unique-id>
    </channel>
  </channel-list>
</adapter>
```

Gives you 
```console
$ curl -si -XPOST -d"Hello World" http://localhost:8080/api/always-fail
HTTP/1.1 500 Server Error
Content-Type: application/json
Transfer-Encoding: chunked

{
  "exceptionLocation" : "ThrowExceptionService(always-fail)",
  "exceptionMessage" : "Always Fails.",
  "workflow" : "always-fail@default-channel"
}
$ curl -si -XGET http://localhost:8080/api/failed/list
HTTP/1.1 200 OK
Content-Type: application/json
Transfer-Encoding: chunked

[{"bucket":"./fs/retries","size":128,"name":"ec24c442-e688-45cc-b6db-24bd6893cf4e","lastModified":1603981015000}]
chanl3@lhrm32445 ~
$ curl -si -XPOST http://localhost:8080/api/retry/ec24c442-e688-45cc-b6db-24bd6893cf4e
HTTP/1.1 202 Accepted
Content-Type: text/plain
Transfer-Encoding: chunked


chanl3@lhrm32445 ~
$
```

Of course the logging will still indicate the the workflow "failed" since that's the configured behaviour but you should see : 
```
TRACE [RetryFromJetty::Retry] [c.a.c.h.j.r.RetryFromJetty.onAdaptrisMessage()] [{}] Attempting to retry ec24c442-e688-45cc-b6db-24bd6893cf4e; resubmitting to [always-fail@default-channel]
```

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/Adapter.java
+++ b/interlok-core/src/main/java/com/adaptris/core/Adapter.java
@@ -174,6 +174,7 @@ public final class Adapter implements StateManagedComponentContainer, ComponentL
     getMessageErrorHandler().registerDigester(getMessageErrorDigester());
     injectErrorHandler();
     registerWorkflowsInRetrier();
+    LifecycleHelper.prepare(getFailedMessageRetrier());
   }
 
   /**
--- a/interlok-core/src/main/java/com/adaptris/core/DefaultFailedMessageRetrier.java
+++ b/interlok-core/src/main/java/com/adaptris/core/DefaultFailedMessageRetrier.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2015 Adaptris Ltd.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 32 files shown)


quotidian-ennui:
Needs to be merged before we can write an AWS / Azure blob variant.

mcwarman (approved):
Confirmed working as per test plan.

quotidian-ennui:
Hmm, I realise that `size` is wrong in the report, since it's reporting the size of the directory as opposed to the size of _payload.blob_. I'm not sure that size makes sense in this instance; so perhaps it should be excluded?

sebastien-belin-adp:
> Hmm, I realise that `size` is wrong in the report, since it's reporting the size of the directory as opposed to the size of _payload.blob_. I'm not sure that size makes sense in this instance; so perhaps it should be excluded?

I should be easy to get the payload.blob size from the message dir for the size.
Size may be useful in some context to decide or not to retry a large message?

PR #562, opened 2020-11-04
Correct dotAll usage for regex in resolve method

albinoloverats:
## Motivation

Discovered that the ternary operator was incorrect when testing the MS Azure/Graph mail consumer/producer.

## Modification

Flipped the results of the ternary operator.

## Result

The resolve now works correctly.

## Testing

I added to the resolve test case my originally non-working example, which now works as expected.


quotidian-ennui (approved):
The first SNYK error is because you probably haven't pulled the latest develop branch (so we can ignore this, since eclipse should be at 9.4.33 or something now).
The 2nd SNYK error is also ignorable, and is down to xerces not using maven central so 2.13.1 isn't available anywhere other than via the redhat/jboss repos (thus the SP3 backport).

PR #563, opened 2020-11-05
INTERLOK-3463 Make the HikariPoolBuilder "lazier"

quotidian-ennui:
## Motivation

If you use the Hikari connection pool via pluggable-jdbc-connnection-pool; and the database is not available, then 
`java -jar lib/interlok-boot.jar -configcheck` doesn't work.

So, if you have this simple configuration, and you don't have a mariaDB instance on localhost running, then -configcheck will fail.

```xml
<adapter>
  <shared-components>
    <connections>
      <pluggable-jdbc-pooled-connection>
        <unique-id>jdbc-pooled-connection</unique-id>
        <test-statement>SELECT DATABASE(), VERSION(), NOW(), USER();</test-statement>
        <driver-imp>com.mysql.jdbc.Driver</driver-imp>
        <always-validate-connection>true</always-validate-connection>
        <username>root</username>
        <password>root</password>
        <connect-url>jdbc:mysql://localhost:3306/mysql?autoReconnect=true</connect-url>
        <builder class="jdbc-hikari-pool-builder"/>
      </pluggable-jdbc-pooled-connection>

    </connections>
    <services/>
  </shared-components>
</adapter>
```

## Modification

- We delay the construction of the HikariDataSource until we actually need to use it (i.e. when a real datasource
method is called). This means that configuration checks which call prepare() don't break if there's no database.
- Refactor PooledDataSourceImpl so that the underlying datasource is provided by the concrete classes.
- Remove the NonNull annotation from PluggableJdbcConnection for the build since it has a sensible default.

## Result

- configcheck now works again with Hikari connection pool

## Testing

- use nightly build, run -configcheck with the above configuration.: __FAILS__
- use nightly build, change the builder element to be jdbc-default-pool-factory : __SUCCESS__
- use nightly build, remove the builder element; run configcheck again : __FAIL__
- checkout the PR, run -configcheck with the above configuration: __SUCCESS__
- checkout the PR, remove the builder element; run configcheck : __SUCCESS__

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/jdbc/C3P0PooledDataSource.java
+++ b/interlok-core/src/main/java/com/adaptris/core/jdbc/C3P0PooledDataSource.java
@@ -4,18 +4,21 @@ import java.io.IOException;
 import com.mchange.v2.c3p0.ComboPooledDataSource;
 
 public class C3P0PooledDataSource extends PooledDataSourceImpl<ComboPooledDataSource> {
+  private ComboPooledDataSource wrapped = null;
 
   public C3P0PooledDataSource(ComboPooledDataSource wrapped) {
-    super(wrapped);
+    super();
+    this.wrapped = wrapped;
   }
 
+  @Override
   public ComboPooledDataSource wrapped() {
     return wrapped;
   }
 
   @Override
   public void close() throws IOException {
-    wrapped.close();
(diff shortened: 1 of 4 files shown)


sebastien-belin-adp (approved):
Changes make sense and I successfully tested the new behaviour.

PR #576, opened 2020-12-03
feature/INTERLOK-2873

albinoloverats:
## Motivation

The metadata comparisons should be usable as operators by ConditionWithOperator classes.

## Modification

The metadata comparators now implement operator, and can be used as such. 

## Result

The users gets some new operators (particularly StartsWith, EndsWith, and Contains, as they already have Equals). They can be used by classes which extend ConditionWithOperator, whilst not breaking backwards compatibility with the MetadataComparisonService.

## Testing

There are new unit tests (that are copied from the original operators) that confirm the new functionality whilst keepin the old. They should now also appear in the UI whenever a ConditionWithOperator wants an operator.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/metadata/compare/ComparatorImpl.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/metadata/compare/ComparatorImpl.java
@@ -28,11 +28,11 @@ public abstract class ComparatorImpl implements MetadataComparator {
   @AutoPopulated
   @Getter
   @Setter
-  protected String resultKey;
+  private String resultKey;
 
   @Getter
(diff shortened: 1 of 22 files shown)


quotidian-ennui (approved):
Fixed typo in the ComponentProfile, but it's good.

aaron-mcgrath-adp:
Yeah, it's good.  Just wondering if the fields in the Impl need to be protected rather than private?

albinoloverats:
> Yeah, it's good. Just wondering if the fields in the Impl need to be protected rather than private?

Good catch - I widened their scope (for whatever reason) while working on it and just didn't narrow it back down again.

PR #579, opened 2020-12-08
INTERLOK-3497 - Add a system property based file report

quotidian-ennui:
## Motivation

To simplify things for the interlok-verify-report out of the parent build.gradle; we need to make the config checker integrate loosely with it.

Since the task works off the back off a file that is emitted via log4j configuration based on the warnings; we can make -configcheck write a file in the same format.

## Modification

- Add a DeprecatedConfigFileReporter that operates based on system properties
- Modify DeprecationConfigChecker so that it includes the unique-id + classname if available.
- Add Tests.

## Result

```
java -Dinterlok.verify.deprecated.filename=./report.txt -jar lib/interlok-boot.jar -configcheck
```

Now emits a report.txt -> if the system property denoting the filename is not present, then no data is written to file.

Available system properties are : 
```
interlok.verify.deprecated.filename
// default is CODE_SMELL
interlok.verify.deprecated.category
// default is INFO
interlok.verify.deprecated.level
```

## Testing

Take this configuration
```xml
<adapter>
  <channel-list>
    <channel>
      <workflow-list>
        <standard-workflow>
          <consumer class="null-message-consumer">
            <destination class="configured-consume-destination">
              <destination>/api/base64</destination>
            </destination>
          </consumer>
            <services>
              <always-fail-service>
               <unique-id>always-fail-cos</unique-id>
              </always-fail-service>
              <service-list>
                <services>
                  <always-fail-service/>
                </services>
              </service-list>
            </services>
          </service-collection>
          <unique-id>clever-edison</unique-id>
        </standard-workflow>
      </workflow-list>
      <unique-id>jovial-feynman</unique-id>
    </channel>
  </channel-list>
</adapter>
```

If you run "gradle check" via the parent gradle system then you get a report.txt file from the parent gradle, this works off the `prepare()` method emitting warnings:

```
$ cat build/reports/interlok/verify/report.txt
CODE_SMELL,INFO,[AlwaysFailService] is deprecated, use [com.adaptris.core.services.exception.ThrowExceptionService] instead
CODE_SMELL,INFO,NullMessageConsumer uses destination, it has no meaning
CODE_SMELL,INFO,[Adapter(MyInterlokInstance)] has a MessageErrorHandler with no behaviour; messages may be discarded upon exception
```

If you build this PR and run it manually : 
```
java -Dinterlok.verify.deprecated.filename=./report.txt -jar lib/interlok-boot.jar -configcheck
```
Then you now get...

```
$ cat ./report.txt
CODE_SMELL,INFO,Interlok Deprecation Warning: [channelList.channels[0].workflowList.workflows[0].consumer.destination][NullMessageConsumer]: Is deprecated. It will be removed in 4.0.0
CODE_SMELL,INFO,Interlok Deprecation Warning: [channelList.channels[0].workflowList.workflows[0].serviceCollection.services[0]][AlwaysFailService(always-fail-cos)]: Is deprecated. It will be removed in a future version
CODE_SMELL,INFO,Interlok Deprecation Warning: [channelList.channels[0].workflowList.workflows[0].serviceCollection.services[1].services[0]][AlwaysFailService]: Is deprecated. It will be removed in a future version
```

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/management/config/DeprecatedConfigFileReporter.java
+++ b/interlok-core/src/main/java/com/adaptris/core/management/config/DeprecatedConfigFileReporter.java
@@ -1,0 +1,58 @@
+package com.adaptris.core.management.config;
+
+import java.io.FileWriter;
+import java.io.PrintWriter;
+import java.util.Collection;
+import org.apache.commons.lang3.StringUtils;
+import lombok.NoArgsConstructor;
+
+/**
+ * Reports on deprecated configuration based on a system property.
+ * <p>
+ * Note that this will only emit output if the system property {@value #SYSPROP_FILENAME} is set as
+ * a system property. It is designed to mimic behaviour of the {@code InterlokVerifyReport} process
+ * made available as part of the parent gradle.
+ * </p>
+ */
+@NoArgsConstructor
+public class DeprecatedConfigFileReporter implements ConfigurationReporter {
+
+  public static final String SYSPROP_FILENAME = "interlok.verify.deprecated.filename";
+  public static final String SYSPROP_CATEGORY = "interlok.verify.deprecated.category";
+  public static final String SYSPROP_LEVEL = "interlok.verify.deprecated.level";
+
+  public static final String CATEGORY_DEFAULT = "CODE_SMELL";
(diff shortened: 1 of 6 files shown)


mcwarman (approved):
Tested and working as expected.

PR #580, opened 2020-12-09
INTERLOK-3526 Report the leaf for javax validation config check

quotidian-ennui:
## Motivation

The changes from https://github.com/adaptris/interlok/pull/579 introduced a leaf node report into when emitting deprecation warnings.

This should also be in the javax validation checker

## Modification

- When running the config checker with javax validation report on the leaf bean that caused the violation.

## Result

If you have a javax validation error; then it will also print out the friendly name (classname + uniqueid)

## Testing

If you have a service badly configured
```
              <http-request-service>
                <unique-id>http-request-service-without-url</unique-id>
              </http-request-service>
```

Then you now get the output : 
```
javax.validation checks:
Failed with exceptions:
Interlok Validation Error: [channelList.channels[0].workflowList.workflows[0].serviceCollection.services[0].url][HttpRequestService(http-request-service-without-url)]: must not be blank
```

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/management/config/DeprecatedConfigurationChecker.java
+++ b/interlok-core/src/main/java/com/adaptris/core/management/config/DeprecatedConfigurationChecker.java
@@ -1,9 +1,6 @@
 package com.adaptris.core.management.config;
 
-import static com.adaptris.core.util.LoggingHelper.filterGuid;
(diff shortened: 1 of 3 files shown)


sebastien-belin-adp (approved):
Yeah, some nice component and unique id in the validation errors.

PR #583, opened 2020-12-10
INTERLOK-3528 xpath Class.forName()

quotidian-ennui:
## Motivation

At the moment whenever you call `new com.adaptris.util.xml.XPath()` there is a call to Class.forName() to figure out which XpathFactory that it really wants to use (this is because Saxon is no longer available via XPathFactory.newInstance()) so we have to figure out which one to use.

The naiive use of Class.forName() can end up blocking Threads, most of the time it doesn't but high thread count + lots of XPaths...
 
```
java.lang.Thread.State: BLOCKED (on object monitor)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:398)
  - waiting to lock <0x00000006c03be418> (a java.lang.Object)
  at org.springframework.boot.loader.LaunchedURLClassLoader.loadClass(LaunchedURLClassLoader.java:92)
  at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
  at java.lang.Class.forName0(Native Method)
  at java.lang.Class.forName(Class.java:264)
  at com.adaptris.util.text.xml.XPath.build(XPath.java:290)
```

## Modification

- Add a static method that does the `Class.forName` once, and then re-use that class instance to create the appropriate XPathFactory if saxon is selected.
- Make the existing "unbounded" map that XmlTransform uses to cache things, a bounded ExpiringMap impl since lets just say that if we're hitting this edge case with XPath, then transforms may well being used as well.
- Remove use of Class.forName() in XmlValidator (used during schema validation) for the same reason... There was never a need to reflectively construct the xerces grammar pool since we'd always fail if it didn't exist; so that's a hard dependency we can't get rid of.

## Result

No visible impact other than the high-thread / xpath use-case.

## Testing

No tests change, no test failures, so no regressions

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/transform/XmlTransformService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/transform/XmlTransformService.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2015 Adaptris Ltd.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


aaron-mcgrath-adp (approved):
Looks fine conceptually.

PR #589, opened 2020-12-17
INTERLOK-218 Remove references/usages of testConnection

albinoloverats:
## Motivation

We are using JDK1.8 and going for 11; a java.sql.Connection#isValid(int) method which should be used to check the validity of the connection has been available since 1.6

Surely by now, all the JDBC drivers will be JDBC4 compatible (we would need to re-visit this if they aren't).

## Modification

Remove all uses of DatabaseConnection member testConnection.

## Result

There is no need for the end user to set a test connection SQL statement.

## Testing

There should no longer be the testConnection field in the UI. The unit tests should still execute as expected. DB connections should still work as expected.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/jdbc/AdvancedJdbcPooledConnection.java
+++ b/interlok-core/src/main/java/com/adaptris/core/jdbc/AdvancedJdbcPooledConnection.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2015 Adaptris Ltd.
- * 
+ *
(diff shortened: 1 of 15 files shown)


quotidian-ennui:
@albinoloverats  can't do this for 3.12.0 since it would break all the config (since you've removed test-statement as a member, and it has a default value so it's explicitly non-null in all configurations).
Suggest you raise a different PR that just marks the member as deprecated (configDeprecated / Removal?) for version 4.0.0 and then we can get rid of it for 4.x

PR #596, opened 2021-01-04
INTERLOK-218 Mark testConnection as deprecated

albinoloverats:
# Motivation

We are using JDK1.8 and going for 11; a java.sql.Connection#isValid(int) method which should be used to check the validity of the connection has been available since 1.6

Surely by now, all the JDBC drivers will be JDBC4 compatible (we would need to re-visit this if they aren't).

See [PR 589](https://github.com/adaptris/interlok/pull/589) for the removal of the field, and the reason why it's only being deprecated for now.

# Modification

Deprecate the testStatement field

# Result

The field should be highlighted as deprecated in the UI.

# Testing

Everything, including unit tests, should work as before. UI testing to confirm deprecation warning.


quotidian-ennui:
- Mark test-statement with ConfigDeprecated instead of Removal (since this means it gets reported on by configcheck).

```
@ConfigDeprecated(removalVersion = "4.0.0", message = "test-statement is superseded by java.sql.Connection#isValid()", groups = Deprecated.class)
``



## JdbcConnection
- Already uses Connection#isValid() to check if the the connection requires a refresh...
    - the catch of AbstractMethodError appears to indicate we've previously expected to be able to just use isValid() but that's failed for a bunch of JDBC drivers...
- alwaysValidateConnection == true; then we execute the JdbcUtil#testConnection

## JdbcPooledConnection

- It's configurable on the "pool" implementation (c3p0 -> has a testConnectionOnCheckout property).
- alwaysValidateConnection == true; then we execute the JdbcUtil#testConnection

Once we remove the test-statement, then we essentially get rid of the execution of JdbcUtil#testConnection(); since it's not available anymore. 

Is that acceptable? (given that isValid() may still fail with ABstractMethodERror?)

quotidian-ennui (approved):

PR #606, opened 2021-01-18
INTERLOK-3561 ThreadContextWorkflow

quotidian-ennui:
## Motivation

From @gerco 

>When running a bunch of PoolingWorkflows with an embedded Jetty connection, I'm noticing various performance issues depending on configuration:
>
>If the jetty executor is set to some number threads and each pooling workflow is set to fewer threads, a situation can occur where one PoolingWorkflow is swamped with work and has dozens of Jetty threads waiting for an available thread in the PoolingWorkflow.
>
>If the Jetty executor is set to the some number of threads and each pooling workflow is set to the same number, we are running many more threads than we ideally would. This is causing issues with dynamic-shared-services caching since these caches can be very large and per-thread so we run out of memory.
>
>Is there a way to run workflows directly on the Jetty threads rather than have a PoolingWorkflow in between, introducing a layer of complexity we don't need?

## Modification

- Refactor a shared hierarchy out of PoolingWorkflow into a "WorkflowWithObjectPool"
- Added a new ThreadContextWorkflow that extends WorkflowWithObjectPool
    - Not synchronized and maintains a configurable object pool to avoid non-thread safe service lists.
- Modify JettyNoBacklogInterceptor to use WorkflowWithObjectPool as its decision point where it previously used PoolingWorkflow.

## Result

A new workflow `ThreadContextWorkflow` is now available to configure.

## Testing

- Have a jetty workflow that contains a standard workflow; the service should just be WaitService (default 20 seconds is fine, but you'll get bored) : `/api/standard-sleep`
- Have a jetty workflow that contains a pooling workflow; the service should be exactly the same as above : `/api/pooled-sleep`
- Test them both with 3 (or more)  simultaneous curl requests and check that behaviour is as expected.
    - the pooled version will all finish in ~20 seconds.
    - The standard version will finish in ~60 seconds (or more), since standardworkflow is synchronized.

- Create a new thread-context-workflow that has the same configuration : `/api/sleep`
    - additional-debug (expert mode) : true (for extra logging).
    - use-current-thread-name : true (this is the default).
- Test the new workflow using simultaneous curl requests...

- This will finish in ~20 seconds proving that we're handling the requests in a threaded fashion that is unrelated to any internal thread pools that we have.

You should have logging that looks something like this, the __qtp*__ is the internal jetty thread pool thread.

```
TRACE [sad-hugle@default-channel(qtp674484384-35)] [c.a.c.ThreadContextWorkflow.logPoolState()] [{}] Current Pool after [Borrow]: Idle [0], Active [1], Max [10]
DEBUG [sad-hugle@default-channel(qtp674484384-35)] [c.a.c.ThreadContextWorkflow.handleMessage()] [{}] start processing msg...
DEBUG [sad-hugle@default-channel(qtp674484384-35)] [c.a.c.ServiceList.applyServices()] [{}] Executing doService on [WaitService(loving-dijkstra)]
TRACE [sad-hugle@default-channel(qtp674484384-37)] [c.a.c.ThreadContextWorkflow.logPoolState()] [{}] Current Pool after [Borrow]: Idle [0], Active [2], Max [10]
DEBUG [sad-hugle@default-channel(qtp674484384-37)] [c.a.c.ThreadContextWorkflow.handleMessage()] [{}] start processing msg...
DEBUG [sad-hugle@default-channel(qtp674484384-37)] [c.a.c.ServiceList.applyServices()] [{}] Executing doService on [WaitService(loving-dijkstra)]
TRACE [sad-hugle@default-channel(qtp674484384-41)] [c.a.c.ThreadContextWorkflow.logPoolState()] [{}] Current Pool after [Borrow]: Idle [0], Active [3], Max [10]
DEBUG [sad-hugle@default-channel(qtp674484384-41)] [c.a.c.ThreadContextWorkflow.handleMessage()] [{}] start processing msg...
DEBUG [sad-hugle@default-channel(qtp674484384-41)] [c.a.c.ServiceList.applyServices()] [{}] Executing doService on [WaitService(loving-dijkstra)]

...
TRACE [sad-hugle@default-channel(qtp674484384-35)] [c.a.c.ThreadContextWorkflow.logPoolState()] [{}] Current Pool after [Return]: Idle [1], Active [2], Max [10]
TRACE [sad-hugle@default-channel(qtp674484384-37)] [c.a.c.ThreadContextWorkflow.logPoolState()] [{}] Current Pool after [Return]: Idle [2], Active [1], Max [10]
TRACE [sad-hugle@default-channel(qtp674484384-41)] [c.a.c.ThreadContextWorkflow.logPoolState()] [{}] Current Pool after [Return]: Idle [3], Active [0], Max [10]
```

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/PoolingWorkflow.java
+++ b/interlok-core/src/main/java/com/adaptris/core/PoolingWorkflow.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2015 Adaptris Ltd.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -16,24 +16,17 @@
 
 package com.adaptris.core;
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Callable;
-import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
-import java.util.function.Consumer;
 import javax.validation.Valid;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 import org.apache.commons.lang3.Range;
-import org.apache.commons.pool2.PooledObject;
-import org.apache.commons.pool2.PooledObjectFactory;
-import org.apache.commons.pool2.impl.DefaultPooledObject;
-import org.apache.commons.pool2.impl.GenericObjectPool;
+import org.apache.commons.pool2.ObjectPool;
 import com.adaptris.annotation.AdapterComponent;
 import com.adaptris.annotation.AdvancedConfig;
 import com.adaptris.annotation.ComponentProfile;
@@ -46,6 +39,8 @@ import com.adaptris.util.FifoMutexLock;
 import com.adaptris.util.NumberUtils;
 import com.adaptris.util.TimeInterval;
 import com.thoughtworks.xstream.annotations.XStreamAlias;
+import lombok.Getter;
+import lombok.Setter;
 
 /**
  * A Workflow that pools ServiceCollections.
@@ -70,10 +65,10 @@ import com.thoughtworks.xstream.annotations.XStreamAlias;
  * <code>AdaptrisMessageConsumer</code> is succesfully stopped will be treated as <b>bad</b> messages and sent directly to the
  * configured {@link com.adaptris.core.ProcessingExceptionHandler}.
  * </p>
- * 
+ *
  * @config pooling-workflow
- * 
- * 
+ *
+ *
  * @author lchan
  * @author $Author: lchan $
  * @see ProcessingExceptionHandler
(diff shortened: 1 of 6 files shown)


sebastien-belin-adp (changes requested):
1) So I've tried the test scenario and that works very nicely. It also looks nice in the UI.

However the one thing I'm not sure about is the use of the JettyNoBacklogInterceptor.

If I limit the number of thread to 2 and call the curl command 3 times, it will still wait for the other threads to finish before it respond with a 503. Is that expected?

2) The other thing not working well, but you guessed it, is the use of quartz polling consumer.
If I poll every 10 seconds with a waiting service of 20 seconds I'll get:

`TRACE [MyWorkflow@Jetty(NonBlockingQuartzThreadPool-38-4)] [QuartzCronPoller] couldn't get lock in run (previous poll has not finished?); waiting for next scheduled poll.`

It does that a few time, I guess until it uses all its threads

3) And last, the javadoc in JettyNoBacklogInterceptor:

`Note that this interceptor only works with {@link PoolingWorkflow};`

should probably be something like:

`Note that this interceptor only works with {@link PoolingWorkflow} and {@link ThreadContextWorkflow};`

quotidian-ennui:
> 1. So I've tried the test scenario and that works very nicely. It also looks nice in the UI.
> 
> However the one thing I'm not sure about is the use of the JettyNoBacklogInterceptor.
> 
> If I limit the number of thread to 2 and call the curl command 3 times, it will still wait for the other threads to finish before it respond with a 503. Is that expected?
> 

Definitely not, I would have expected a 503 on the 3rd try immediately; so that needs to be fixed.
I think that NoBacklog may not work anyway if I think about it properly.
- If you're in the jetty thread; you haven't handed off to anything to do the work.
- workflowStart() in the interceptor checks the count of messages in flight vs the number of messages in the pool (2=2), so it attempts to do the sendRedirect(503).
- I suspect that it might be the case because you're in the jetty thread you don't actually send the 503 until you manage to do the pool.borrow() and process it.

The key difference is that PoolingWorkflow does the workflowStart() check "in the jetty thread", but then immediately does a `threadpool.submit()`  which returns immediately such that the sendRedirect(503) can effectively "happen".

The fix-it-now solution is to make jetty no backlog only work with PoolingWorkflow (which removes (3)). In retrospect, it may be that if you're going to tune jetty, then you should tune jetty, and jetty-no-backlog is irrelevant to you, since you can probably configure the 503 behaviour in jetty.xml 

> 1. The other thing not working well, but you guessed it, is the use of quartz polling consumer.
>    If I poll every 10 seconds with a waiting service of 20 seconds I'll get:
> 
> `TRACE [MyWorkflow@Jetty(NonBlockingQuartzThreadPool-38-4)] [QuartzCronPoller] couldn't get lock in run (previous poll has not finished?); waiting for next scheduled poll.`
> 

Do we eventually return the object to the object pool so that polling starts working, or does it just fail forever once the pool is exhausted?

sebastien-belin-adp (approved):
JettyNoBacklogInterceptor now only supports PoolingWorkflow
A new issue will be raised for the QuartzCronPoller + ThreadContextWorkflow issue

PR #610, opened 2021-01-25
INTERLOK-3572 NPE on requestor timeout.

aaron-mcgrath-adp:
## Motivation

If you have a standalone requestor that sends to a JMS queue and then waits for a response on another queue, but times out rather than returning a message you get an NPE.  The unit tests show that we trap the exception and handle accordingly, but it's still ugly to have an NPE.

## Modification

Both the JmsProducer and DefinedJmsProducer in the request method attempted to translate any response fromm the reply queue.  Of course we might not get a response hence the NPE.  I've simply added a null check to the response object and throw the appropriate exception when it's null and when it's not null we continue as normal. 

**EDIT:**
The fix has now been moved into the translator itself, such that if you pass in a null message, you get a null result which gives ends in the correct behaviour.

## Result

In terms of functionality nothing really changes except the fact you won't have an ugly NPE stack trace in your logs when a requestor times out.

## Testing

See the enclosed sample solace config.  It will connect to a localhost Solace instance on the default port of 55555 with the default admin user and password.  It listens for messages on "Sample.Q1" queue and will then use a standalone producer to send to a topic named "Sample.T1" waiting for a response, which will never come and timeout after 10 seconds.

Without this fix you'll see an NPE in the logs.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/jms/DefinedJmsProducer.java
+++ b/interlok-core/src/main/java/com/adaptris/core/jms/DefinedJmsProducer.java
@@ -128,10 +128,10 @@ public abstract class DefinedJmsProducer extends JmsProducerImpl {
       receiver = currentSession().createConsumer(replyTo);
       doProduce(msg, dest, replyTo);
       Message jmsReply = receiver.receive(timeout);
-      if(jmsReply == null)
-        throw new JMSException("No Reply Received within " + timeout + "ms");
-      
-      translatedReply = MessageTypeTranslatorImp.translate(getMessageTranslator(), jmsReply);
+      translatedReply =
+          Optional.ofNullable(MessageTypeTranslatorImp.translate(getMessageTranslator(), jmsReply))
+              .orElseThrow(
+                  () -> new JMSException("No Reply Received within " + timeout + "ms"));
       acknowledge(jmsReply);
       // BUG#915
       commit();
--- a/interlok-core/src/main/java/com/adaptris/core/jms/DefinedJmsProducer.java
+++ b/interlok-core/src/main/java/com/adaptris/core/jms/DefinedJmsProducer.java
@@ -128,10 +128,10 @@ public abstract class DefinedJmsProducer extends JmsProducerImpl {
       receiver = currentSession().createConsumer(replyTo);
       doProduce(msg, dest, replyTo);
       Message jmsReply = receiver.receive(timeout);
-      translatedReply =
-          Optional.ofNullable(MessageTypeTranslatorImp.translate(getMessageTranslator(), jmsReply))
-              .orElseThrow(
-                  () -> new JMSException("No Reply Received within " + timeout + "ms"));
+      if(jmsReply == null)
+        throw new JMSException("No Reply Received within " + timeout + "ms");
+      
+      translatedReply = MessageTypeTranslatorImp.translate(getMessageTranslator(), jmsReply);
       acknowledge(jmsReply);
       // BUG#915
       commit();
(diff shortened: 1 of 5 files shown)


aaron-mcgrath-adp:
`<adapter>
  <start-up-event-imp>com.adaptris.core.event.StandardAdapterStartUpEvent</start-up-event-imp>
  <heartbeat-event-imp>com.adaptris.core.HeartbeatEvent</heartbeat-event-imp>
  <shared-components>
    <connections>
      <jms-connection>
        <unique-id>solace</unique-id>
        <user-name>admin</user-name>
        <password>admin</password>
        <vendor-implementation class="basic-solace-implementation">
          <broker-url>tcp://localhost:55555</broker-url>
          <message-vpn>default</message-vpn>
        </vendor-implementation>
      </jms-connection>
    </connections>
    <services/>
  </shared-components>
  <event-handler class="default-event-handler">
    <unique-id>DefaultEventHandler</unique-id>
    <connection class="null-connection">
      <unique-id>determined-shockley</unique-id>
    </connection>
    <producer class="null-message-producer">
      <unique-id>kickass-nightingale</unique-id>
    </producer>
  </event-handler>
  <heartbeat-event-interval>
    <unit>SECONDS</unit>
    <interval>30</interval>
  </heartbeat-event-interval>
  <message-error-handler class="null-processing-exception-handler">
    <unique-id>condescending-bohr</unique-id>
  </message-error-handler>
  <failed-message-retrier class="no-retries">
    <unique-id>goofy-leakey</unique-id>
  </failed-message-retrier>
  <channel-list>
    <channel>
        <lookup-name>solace</lookup-name>
      </consume-connection>
      <produce-connection class="null-connection"/>
      <workflow-list>
        <standard-workflow>
          <consumer class="jms-consumer">
            <unique-id>naughty-wing</unique-id>
            <acknowledge-mode>CLIENT_ACKNOWLEDGE</acknowledge-mode>
            <message-translator class="auto-convert-message-translator">
              <jms-output-type>Text</jms-output-type>
            </message-translator>
            <endpoint>jms:queue:Sample.Q1</endpoint>
          </consumer>
            <unique-id>thirsty-bohr</unique-id>
            <services>
              <log-message-service>
                <unique-id>loving-clarke</unique-id>
                <log-level>DEBUG</log-level>
              </log-message-service>
              <standalone-requestor>
                <unique-id>naughty-keller</unique-id>
                  <lookup-name>solace</lookup-name>
                </connection>
                <producer class="jms-producer">
                  <unique-id>sleepy-leakey</unique-id>
                  <acknowledge-mode>CLIENT_ACKNOWLEDGE</acknowledge-mode>
                  <message-translator class="text-message-translator"/>
                  <delivery-mode>PERSISTENT</delivery-mode>
                  <session-factory class="jms-default-producer-session"/>
                  <endpoint>jms:topic:Sample.T1</endpoint>
                </producer>
                <reply-timeout>
                  <unit>SECONDS</unit>
                  <interval>10</interval>
                </reply-timeout>
              </standalone-requestor>
            </services>
          </service-collection>
          <producer class="null-message-producer">
            <unique-id>distracted-lumiere</unique-id>
          </producer>
          <produce-exception-handler class="null-produce-exception-handler"/>
          <unique-id>zen-swartz</unique-id>
          <message-metrics-interceptor>
            <unique-id>zen-swartz-MessageMetrics</unique-id>
            <timeslice-duration>
              <unit>MINUTES</unit>
              <interval>5</interval>
            </timeslice-duration>
            <timeslice-history-count>12</timeslice-history-count>
          </message-metrics-interceptor>
          <in-flight-workflow-interceptor>
            <unique-id>zen-swartz-InFlight</unique-id>
          </in-flight-workflow-interceptor>
        </standard-workflow>
      </workflow-list>
      <unique-id>nostalgic-hoover</unique-id>
      <auto-start>false</auto-start>
    </channel>
  </channel-list>
  <message-error-digester class="standard-message-error-digester">
    <unique-id>ErrorDigest</unique-id>
    <digest-max-size>100</digest-max-size>
  </message-error-digester>
</adapter>
`

quotidian-ennui:
Probably worth writing a test just for the method. just to avoid code-cov... since it's static it shouldn't be awful to do mock if needs be (or use ActiveMqTextTranslator or whatever).

sebastien-belin-adp (approved):
The new test cover the change in MessageTypeTranslatorImp even though codecov still complains about it (which I'm not sure why).

PR #611, opened 2021-01-26
Remove the tag type from docs

albinoloverats:
Remove the tag type (in this instance @config) from the JavaDocs.


quotidian-ennui:
Well doesn't that suck, I need to update the github action (!)

quotidian-ennui (approved):

PR #621, opened 2021-02-05
feature/interlok-3567 resolveObject

albinoloverats:
## Motivation

With the removal of consumer/producer destinations, a replacement is necessary.

## Modification

Add new method to retrieve an object from object metadata using a similar method to the existing resolve() method

## Result

Objects in object headers/metadata can be retrieve using %messageObject{key}

## Testing

Objects in metadata should be retrieved as their objects (not String) from Object headers.


sebastien-belin-adp:
@albinoloverats I can't really see where resolveObject is going to be called and where %messageObject is going to be used?
Otherwise the code looks good.
I would put "^.*%messageObject\\{([\\w!\\$\"#&%'\\*\\+,\\-\\.:=]+)\\}.*$" into a constant like RESOLVE_REGEXP just because it's cleaner.

sebastien-belin-adp:
@albinoloverats Are you still working on it?
Otherwise could you give us an example on how we can test it?
Where can we use %messageObject{key} in a config?
Thanks

albinoloverats:
@sebastien-belin-adp Yeah, I'm done with this, unless there's anything else that anyone else thinks is necessary. I'll put together some config for testing.

albinoloverats:
Simple config that consumes from a queue and then produces it to whatever `Reply To` is set to - uses a bare install of ActiveMQ with 2 queues: `jms:queue:TestQ1` and `jms:queue:TestQ2`.

[adapter.xml.zip](https://github.com/adaptris/interlok/files/5997112/adapter.xml.zip)

sebastien-belin-adp:
@albinoloverats I've made a small commit mostly removing unused import and warnings.

Other than that I've tested the config and it works nicely with the new changes.

But I have a few questions:

- In JmsProducer line 230 (https://github.com/adaptris/interlok/pull/621/files#diff-f283b4edc0339a3d1be3f531b9af687482fd4a0847343ead4971a90dbd43932eR230) Are we happy with the String cast? Is that never going to cause some class cast exception as resovleObject can return an object? There are a couple more of String cast in PasProducer and PtpProducer.

- And that brings us to JmsProducer line 137 (https://github.com/adaptris/interlok/pull/621/files#diff-f283b4edc0339a3d1be3f531b9af687482fd4a0847343ead4971a90dbd43932eR137):  `Object jmsDest = endpoint(msg);` But jmsDest will never be anything else than a String. Which means that the following check is never going to be true:
`if (jmsDest instanceof Destination) {`

- In DefinedJmsProducer line 59 (https://github.com/adaptris/interlok/pull/621/files#diff-dcbb55f23b37e7fe190b868aafe2427395511e6d340257639b13eb5793cff4cfR59), do we need this check: `if (!JMS_ASYNC_STATIC_REPLY_TO.equals(o)) {`?

- Wouldn't be nice to have some javadoc in InterlokMessage#resolveObject like for InterlokMessage#resolve?

sebastien-belin-adp (approved):
I'm happy with the changes

PR #635, opened 2021-02-22
feature/interlok-3566 - Remove the deprecated produce-destination/consume-destination

albinoloverats:
**Although this is only a draft and there might be things missing, I am looking for feedback to make sure it's on the right track.**

Particularly, I know that `ProduceDestination` and `ConsumeDestination` still exist (for now).

## Motivation

Fully remove the deprecated produce/consume destinations from the various consumer/producer components. Calls to `DestinationHelper` has essentially been inlined as necessary.

## Modification

Remove usages of `ConsumeDestination` and `ProduceDestination`. There are likely more that I've missed and will continue to update as I find them (hence this is only a draft for now).

## Result

They cannot use `setDestination(Destination)` but must now use alternatives (endpoint/topic/queue/etc.)

## Testing

As the UI doesn't display deprecated options by default, an older config that actually uses Destinations will be necessary. This will then fail to load as the ability to `setDestination` no longer exists - *it works by not working*. Beyond that, existing configs that already use the replacements to destinations should continue to work as expected.


albinoloverats:
Let's see what we think - there's probably still a fair bit to do. Either I've culled too much or not enough. Codecov clearly doesn't like me though as I'm getting different results with Jacoco.

interlok-core/src/main/java/com/adaptris/core/AdaptrisMessageImp.java
```diff
@@ -413,18 +413,22 @@ private String resolve(String s, Pattern pattern) {
    */
   @Override
   public Object resolveObject(String s) {
+    if (s == null) {
```
sebastien-belin-adp, line 416: @albinoloverats it's probably better to check all the test coverage warnings and add tests where appropriated such as for AdaptrisMessageImp#resolveObject for line 416 and 417.

interlok-core/src/test/java/com/adaptris/interlok/junit/scaffolding/jms/MockProducer.java
```diff
@@ -10,6 +10,10 @@
 import com.adaptris.core.jms.DefinedJmsProducer;
 import com.adaptris.core.jms.ProducerSession;
 
+import javax.jms.Destination;
```
sebastien-belin-adp, line 13: Line 13, 14, 15, imports added for no reason.

sebastien-belin-adp (changes requested):
## Main Code

- AdaptrisMessageConsumerImp: Do we have tests for "newThreadName()", with listener  being null and non null?

- AdaptrisMessageImp: Added line #L417 is not covered by tests.

- MarshalledClassDestination: Is it ok to remove it? Do we still need some similar behaviour?

- AggregatingFsConsumer: Remove some white space line, and fix warning for AdaptrisMessageEncoder.

- FsConsumerImpl: "destinationWarningLogged" is no longer needed (Line 187).
Also do we want to keep "protected void prepareConsumer() throws CoreException" as two of the three class that extends FsConsumerImpl delcared an empty prepareConsumer() as well.
If we keep it, we may want to remove it in TraversingFsConsumer and NonDeletingFsConsumer.

- AggregatingFtpConsumer: Fix warning for AdaptrisMessageEncoder.

- FtpConsumerImpl: "destinationWarningLogged" is no longer needed (Line 119). Unused import import com.adaptris.core.AdaptrisMessageListener; (Line 23).

- FtpProducer: "destWarning" and "requestReplyWarning" are no longer needed (Line 181, 182). There are also other warning concerning derecated things but that probably don't need to be done in this task.

- BasicJettyConsumer: "destinationWarningLogged" is no longer needed (Line 88).

- JdbcStoredProcedureProducer: Added line #L354 is not covered by tests.

- AggregatingQueueConsumer: Added lines #L78 - L79 are not covered by tests.

- JmsConsumer: Remove "destination" from @DisplayOrder annotation. (Probably need to do that in many other consumers and producers).

- JmsPollingConsumerImpl: Added line #L97 are not covered by tests. We may get rid of abstract "configuredEndpoint()" and make "endpoint()" abstract?

- PasConsumer: Added line #L104 is not covered by tests. Unused import import com.adaptris.core.AdaptrisMessageListener; (Line 21).

- JmxNotificationConsumer: Fix typo in javadoc for "notificaiton" should be "notification" (Line 177) and "notifcation" should be "notification" (Line 248).

- AggregatingConsumerImpl: Still using some sort of destination. Should we deprecated that and use endpoint (or whatever appropriated name) and filterExpression like in other consumers?

- Global: Shall we support expressions for all the "destinations" (endpoint, urls, ...)? Maybe in another task?

- MessageDrivenDestination: Should we deprecated that and use endpoint (or whatever appropriated name)? Maybe in another task?

## Tests

- StandaloneProducerTest: We should remove "testDoProduceWithDest()" as it does not make sense anymore and does the same as "testDoProduce()"

- FtpCase: "testBasicConsumeWithOverride()" should use "getDestinationStringWithOverride()" instead of "getDestinationString()" when setting endpoint?
"testConsumeWithFilter()" should set the fileFilter as well as before it was using "new ConfiguredConsumeDestination(getDestinationString(), "*.txt")".
"testConsumeWithNonMatchingFilter()" should set the fileFilter as well as before it was using "new ConfiguredConsumeDestination(getDestinationString(), "*.xml")".
Remove the "@SuppressWarnings("deprecation")" that are not needed. Fix other warnings.

- RelaxedFtpCase: "testConsume_OverrideUrl()" should use "getDestinationStringWithOverride()" instead of "getDestinationString()" when setting endpoint?
"testConsumeWithFilter()"s hould set the fileFilter as well as before it was using "new ConfiguredConsumeDestination(getDestinationString(), "*.txt")".
"testConsumeWithNonMatchingFilter()" should set the fileFilter as well as before it was using "new ConfiguredConsumeDestination(getDestinationString(), "*.xml")".
Remove the "@SuppressWarnings("deprecation")" that are not needed. Fix other warnings.

- RelaxedFtpConsumerTest: In "testWithIncorrectFilterConsume()" "consumer.setFileFilterImp("myFilter");" is not needed (Line 311).
Remove the "@SuppressWarnings("deprecation")" that are not needed. Fix other warnings.

- RelaxedFtpProducerTest: Remove the use of ConfiguredProduceDestination "createForTests(ConfiguredProduceDestination dest)" could be "createForTests(String dest)" instead (Line 142).
Remove the "@SuppressWarnings("deprecation")" that are not needed.

- LargeFsMessageProducerTest: "testProduceWithOverride()" should do "producer.setBaseDirectoryUrl(PROPERTIES.getProperty(BASE_KEY));" and remove "//new ConfiguredProduceDestination(PROPERTIES.getProperty(BASE_KEY)));" (Line 131, 132).
Remove the "@SuppressWarnings("deprecation")" that are not needed. Fix other warnings.

- ExampleProduceDestinationCase: Maybe "protected abstract Object createDestinationForExamples()" should be "protected abstract MessageDrivenDestination createDestinationForExamples()" instead? And child classes updated accrdingly.

- MockProducer: Remove added unused imports.

## Nice to do but not a priority

- AdapterTest: Remove construtor, not needed. Remove the "@SuppressWarnings("deprecation")" that is not needed.

- XStreamMarshallerTest: Remove "XSTREAM_STANDARD_XML" as its not used anymwhere (Line 48).

- FileInputStreamDataInputParameterTest, FsConsumerCase, FsImmediateEventPollerTest, FsMessageConsumerTest, FsMessageProducerTest, MovingDeletingFsConsumerTest, TraversingFsConsumerTest, FtpConsumerCase, FtpConsumerTest, EmbeddedHttpConsumerTest:
Remove the "@SuppressWarnings("deprecation")" and fix warning.

- NullMessageProducerTest, ProduceOnlyProducerImpTest, QuartzCronPollerTest, RequestReplyProducerTest: Remove the "@SuppressWarnings("deprecation")" that are not needed. Fix other generics warnings.

- EmbeddedJettyHelper: Unused import import com.adaptris.core.ConfiguredProduceDestination; (Line 22).

- JmsProducerTest, PasProducerTest, PtpProducerTest, MockNonStandardRequestReplyProducer: Fix warnings.

albinoloverats:
> AdaptrisMessageConsumerImp: Do we have tests for "newThreadName()", with listener being null and non null?

The `NullMessageConsumerTest` (as far as I can tell) doesn't set a listener so it'll be null in that instance, although I don't see anything else that does set one.

> MarshalledClassDestination: Is it ok to remove it? Do we still need some similar behaviour?

This could be one of the times I went too far removing destinations - it can always be added back if necessary.

> FsConsumerImpl: "destinationWarningLogged" is no longer needed (Line 187).
Also do we want to keep "protected void prepareConsumer() throws CoreException" as two of the three class that extends FsConsumerImpl delcared an empty prepareConsumer() as well.
If we keep it, we may want to remove it in TraversingFsConsumer and NonDeletingFsConsumer.

`AdaptrisPollingConsumer#prepareConsumer` could be made non-abstract and then all of the empty implementations could be removed. That would leave `PollingTrigger.prepareConsumer` to do its own thing. For now I've just removed the method from `TraversingFsConsumer` and `NonDeletingFsConsumer`.

> AggregatingConsumerImpl: Still using some sort of destination. Should we deprecated that and use endpoint (or whatever appropriated name) and filterExpression like in other consumers?

I think this is pretty much what it's doing, the `ConsumeDestinationGenerator` is just an `String` endpoint and filter expression. But similarly to the usages of `MessageDrivenDestination`, they've not been deprecated so I was hesitant to just blindly remove them.

Time to move on the the tests...

higgyfella:
I built a basic Interlok instance with no optionals, and the jars assembled from this branch.

I then tested all the consumers and producers I could find, within the UI, tested, as in, I tested the configuration of them using the UI, looking at the form and xml view, and then the normal, advanced, expert view. I found no issues.

classes tested in the ui include:
- NullMessageConsumer
- PollingTrigger
- FsConsumer
- JmsPollingConsumer
- PtpPollingConsumer
- PasPollingConsumer
- LargeFsConsumer
- MovingNonDeletingFsConsumer
- NonDeletingFsConsumer
- TraversingFsConsumer
- FtpConsumer
- JettyMessageConsumer
- JmsConsumer
- PtpConsumer
- JmsSyncConsumer
- PasConsumer
- JmxNotificationConsumer
- RelaxedFtpConsumer
- StandaloneConsumer
- NullMessageProducer
- RetryOnceStandaloneProducer
- StandaloneProducer
- FsProducer
- StandardResponseProducer
- LargeFsProducer
- StandardHttpProducer
- FtpProducer
- JdbcStoredProcedureProducer
- JmsAsyncProducer
- JmsProducer
- PtpProducer
- PasProducer
- RelaxedFtpProducer

sebastien-belin-adp:
> the jmx-notification-consumer should have the 'object-name' field marked as a required field.

That worth checking all of them to make sure we have @NotBlank or @NotNull for all the new "destinations" when it makes sense.

aaron-mcgrath-adp (approved):
It's going to need a thorough test through the optionals.... but here we go.

higgyfella (approved):
happy to approve, looked good in ui, understood that it'll break optionals

PR #640, opened 2021-03-02
INTERLOK-3439 Log a warning when a javascript name is used with a scripting service

sebastien-belin-adp:
## Motivation

Nashorn script engine is deprecated and will be removed in Java 17 so we need to warn users of ScriptingService and EmbeddedScriptingService and propose an alternative.

## Modification

- When using a scripting service the code now check if the script language is a nashorn supported language("js", "javascript"...)  and log a warning if it is and GraalJS is not in the classpath.
- If the script language is "nashorn" we log a warning regardless if GraalJS is in the classpath or not.
- Added new tests

Graaljs supports the all the javascript names Nashorn supports ("js", "JS", "JavaScript", "javascript", "ECMAScript", "ecmascript") so we don't need to do any translation.

The adapter need to be started with a system property so GraalJS works the same way as Nashorn:
`polyglot.js.nashorn-compat=true`

In the attached project it's set up in the bootsrap.properties.

## Result

- The scripting service still works the same if not using GraalJS but the user will get a warning.
- The scripting service will still works the same if using GraalJS with a javascript language ("javascript", "js", "ecmascript"...).
- JRuby is not affected

## Testing

- Use the attached project [graaljs.zip](https://github.com/adaptris/interlok/files/6065477/graaljs.zip) and build it: `gradlew clean assemble`
- Replace interlok-core.jar with your build from this PR branch.
- Start the adapter `(cd ./build/distribution && java -jar lib/interlok-boot.jar)`

The config has a pulling trigger that create a message every 10s and an EmbeddedScriptingService that add a payload to a message.

- With no change to the project you should not get a any warning.
- If you change the EmbeddedScriptingService  language to "nashorn" and restart the adapter you should get a warning about Nashorn being deprecated. In fact two warnings as there is the warning from Nashorn and the warning from the scripting service.
- If you change back the EmbeddedScriptingService  language to "javascript" and remove the graaljs dependencies graal-sdk.jar, js.jar and js-scriptengine.jar and restart the adapter you should also get the two warnings.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/ScriptingServiceImp.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/ScriptingServiceImp.java
@@ -1,28 +1,34 @@
 /*
  * Copyright 2015 Adaptris Ltd.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  * See the License for the specific language governing permissions and
  * limitations under the License.
-*/
+ */
 
 package com.adaptris.core.services;
 
 import java.io.IOException;
 import java.io.Reader;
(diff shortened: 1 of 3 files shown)


quotidian-ennui (approved):
We should probably have documentation as to how to change build.gradle to include the graalljs dependencies in a vanilla JVM.

sebastien-belin-adp:
> We should probably have documentation as to how to change build.gradle to include the graalljs dependencies in a vanilla JVM.

I'll add a page or a section for this scripting service in the doc so I can add that with it.

PR #648, opened 2021-03-09
INTERLOK-3573 XML services should use a restricted instance to mitigate against XXE

quotidian-ennui:
## Motivation

XpathMetadataService is using DocumentBuilderFactoryBuilder.newInstanceIfNull, which relies on the underlying document factory for things like XXE mitigations. It should probably be using newRestrictedInstanceIfNull instead to explicitly disable things like doctype and entity handling.

On the same note, DocumentBuilderFactoryBuilder should be setting those same features to "off" when not configured rather than rely on the default underlying DocumentBuilderFactory configuration.

This change will not break config directly, but it will break currently-working functionality if someone was relying on those featured being defaulted on in some DocumentBuilderFactory implementations.

## Modification

- newInstance() now just delegates to newRestrictedInstance()
- Modify enums so that the new "secure" defaults are applied before any values taken from configuration.
- Add a newLenientInstance/ifNull for completeness.
- Add Tests.

## Result

There is no configuration change for the end-user however there is a massive behaviour change which has knock-on effects.

- expand-entity-references = false by default.
- `http://apache.org/xml/features/disallow-doctype-decl` is set to true as the default feature.
- ALL XML Services are affected that end up having a configurable DocumentBuilderFactoryBuilder (xml-transform / xpath-metadata etc etc).

 
## Testing

Configure an XPathMetadataService that doesn't have a DocumentBuilderFactoryBuilder configured : 

e.g. 
```
<xpath-metadata-service>
   <xpath-query class="configured-xpath-query">
     <metadata-key>adapter_unique_id</metadata-key>
     <xpath-query>/adapter/unique-id</xpath-query>
   </xpath-query>
</xpath-metadata-service>
```

Along with a corresponding input file.

```
<!DOCTYPE adapter [
<!ENTITY MY_ADAPTER_ID 'my-adapter'>

]>

<adapter>
  <unique-id>&MY_ADAPTER_ID;</unique-id>
</adapter>
```

- This PR should fail with an exception containing something like `DOCTYP not allowed`
- The 4.0-SNAPSHOT branch should succeed.

Note that this is very similar to the test contained in XpathMetadataServiceTest#testDoService_DocType_DefaultBehaviour

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/util/DocumentBuilderFactoryBuilder.java
+++ b/interlok-core/src/main/java/com/adaptris/core/util/DocumentBuilderFactoryBuilder.java
@@ -13,6 +13,7 @@ import org.xml.sax.EntityResolver;
 import com.adaptris.annotation.AdvancedConfig;
 import com.adaptris.annotation.AutoPopulated;
 import com.adaptris.annotation.DisplayOrder;
+import com.adaptris.annotation.InputFieldDefault;
 import com.adaptris.util.KeyValuePair;
 import com.adaptris.util.KeyValuePairSet;
 import com.thoughtworks.xstream.annotations.XStreamAlias;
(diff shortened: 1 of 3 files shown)


sebastien-belin-adp (approved):
Although it's not part of this changes, the constant DISABLE_DOCTYP has a typo, should be DISABLE_DOCTYPE?
Otherwise everything works fine.

quotidian-ennui:
> Although it's not part of this changes, the constant DISABLE_DOCTYP has a typo, should be DISABLE_DOCTYPE?
> Otherwise everything works fine.

I suspect it's a vestigal remnant of SAP influece (since they seem to use IDOCTYP); no reason not to change it other than ruining peoples day cos it's a public constant (it's not used anywhere other than that class)

PR #649, opened 2021-03-09
Attempt to retry tests for intermittent failures

quotidian-ennui:
## Motivation

Some junit tests appear to be sensitive in the sense that they cannot be re-created on "real" machines but are intermittently failing when running in virtual environments like the various CI providers we use.

For instance:
- jmx-notifications sometimes fail
- jdbc pooled connections sometimes fail.

What we need to do is to be able to retry tests if they fail, to make sure that they aren't intermittent ones

## Modification

- apply the org.gradle.test-retry plugin
- Configure the retry plugin in interlok-core to only trigger when we are in CI mode (github actions, appveyor, circleci will auto set CI=true as an env-var).
- Encapsulate some existing "functions" into a pseudo module, rather than using `buildInfo()` we use `buildDetails.buildInfo()` etc. It's just for namespacing as our gradle files get more awesome.

## Result

No change for the user running on their desktop/laptop whatever

## Testing

Well, let's check the builds that are triggered as part of this PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,6 +6,7 @@ plugins {
   id 'org.owasp.dependencycheck' version '6.1.1' apply false
   id "io.freefair.lombok" version "5.3.0" apply false
   // id 'org.kordamp.gradle.jdeps' version '0.12.0' apply false
+  id "org.gradle.test-retry" version "1.2.0" apply false
 }
 
 ext {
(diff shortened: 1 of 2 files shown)


sebastien-belin-adp (approved):
Hard to see if that's working as no test failed this time.
But the gradle changes look good.

PR #656, opened 2021-03-16
bug/interlok-3748 Bring back the original bahaviour of FileParameter

albinoloverats:
## Motivation

The changes to `FileParameter` meant it no longer had the same behaviour as `Destination`.

## Modification

* `resolveObject` no longer calls the `resolve` method
* Use `resolve` instead of `resolveObject` where applicable
* Add `InputFieldHint` annotation to endpoint field
* Update unit tests

## Result

That `FileParameter` should behave as before the removal of the `Destination` classes.

## Testing

Existing configurations that use `FileParameter` should work as expected.


quotidian-ennui:
Can't approve this since there's some impact from the work being done for INTERLOK-3559 (it'll just be safe to merge develop back into this branch since that has a load more changes, and JMS is affected).

quotidian-ennui:
You changed the semantics of resolveObject() which means you have to think about *everywhere you use that method*

- HttpProducer no longer works; since `%message{XXXX}` will not resolve as a valid endpoint.
- DefinedJmsProducer no longer works; for *exactly the same reason*. It needs to support resolveObject but then choose to use resolve() if the object doesn't exist in object metadata.
```
      Object o = msg.resolveObject(JMS_ASYNC_STATIC_REPLY_TO);
      Destination replyTo = null;
      if (o instanceof Destination) {
        replyTo = (Destination)o;
      } 
      if (replyTo == null) {
         if (msg.containsKey(JMS_ASYNC_STATIC_REPLY_TO) {
            replyTo = createDestinatoin(msg.getMetadataValue(JMS_ASYNC_STATIC_REPLY_TO));
         }
      }
```
- PasProducer and PtpProducer no longer work for the same reason.
- JmsProducer doesn't work for the same reason.


YOu need to enable the JMS tests, and write tests to explicitly test the branches where we attempt to do JMS producing with

- %messageObject{JMSReplyTo}
- %messageObject{SomeOtherObject that is a Destination}
- %messageObject{SomeObject that is just an object}
- %message{myQueueName} // where MyQueue  points to is either a queue, topic, or jms:topic:topicName 
- "myQueue"

quotidian-ennui (changes requested):
See comment above.

quotidian-ennui:
JmsProducer is still probably broken, if you follow the path through for

- %message{MyKey}
- %messageObject{JMSReplyTo}
- %messageObject{someRandomKey}
- jms:topic:myTopic

What's the java.jms.Topic associated with JmsDestination ?

```
  protected JmsDestination buildDestination(AdaptrisMessage msg, boolean createReplyTo)
```

If I run it through my head

- %message{MyKey} -> probably throws an Exception since the string isn't in the format "jms:topic:XXX" failing on line 147
- %messageObject{JMSReply} -> if JMSReplyTo exists in object metadata and is a destination then we use that; otherwise it will probably fail on line 147 if it doesn't exist, or if it does and isn't a destination it will fail on line 143
- jms:topic:myTopic -> JmsDestination wraps a javax.jms.Topic whose `name` is myTopic...

albinoloverats:
Superseded by #659

PR #657, opened 2021-03-16
INTERLOK-3559 Remove things marked for removal in 4.0.0

quotidian-ennui:
## Motivation

Remove things that are marked for removal in 4.0.0 since 4.0.0 is considered a breaking change, we can remove a bunch of configuration that we're carrying around.

This PR should serve as the baseline 

## Modification

Categorised into what I think are high/med/low impact changes, where high is equivalent to "we know it's being used and there will be config pain"...

### High Impact Changes

- Remove tokenKeys from GetOauthToken since we have AccessTokenWriter (OAUTH users who don't use the UI).
- Remove `durable` from PasConsumer ( just remove it)
- Remove PayloadFromMetadataService 
- Remove com.adaptris.core.util.CloseableIterable (_high impact_; breaks downstream optional packages)
- Removed SplitJoinService + PoolingSplitJoinService (switch to using pooled-split-join-service).
- Removed test-statement from database connection : could break a bunch of stuffs this supersedes https://github.com/adaptris/interlok/pull/589

### Medium Impact 

- Promote "condition" in JettyRouteSpec by removing other configuration (@mcwarman et. al.)
- Remove XmlSchemaValidator (replacements in com.adaptris.core.transform.schema)
- Removed some Conversion methods, and switch to proxying java.util.Base64 where possible. Methods remain deprecated, downstream projects might be broken but only from a "code perspective"

### Low Impact

- Remove key/value from LoggingContextWorkflowInterceptor
- Remove MetadataOutputStreamWrapper
- Removed LogHandler and associated items : possible impact on people who might use the runtime JMX component associated with FileLogHandler, UI ?

## Result

There's a bunch of stuff that's no longer available to config.
Some adaptris-labs projects may be broken.

## Testing

The modifications detail what's changed, and I have tried to keep each commit only related to a single "suite" of changes.
Need to review all the test runtime projects.


higgyfella (approved):
I looked at the code changes, and there's a lot, but I couldn't find any issues to raise; I also checked the tests pass locally, which they did pass; then I used the jars i build from this branch in the ui, and everything started okay.

So I'm happy to approve, and we can get on with finding downstream issues.

PR #659, opened 2021-03-18
INTERLOK-3748 + INTERLOK-3567 Change semantics of resolveObject / JMSReplyToDestination

quotidian-ennui:
## Motivation

resolveObject semantics needs to change so that it's more "unix-philosophy-alike". This change also removes use of resolveObject where things are expected to be Strings anyway.

## Modification

- Change semantics of resolveObject so it doesn't check normal metadata.
- Remove use of resolveObject from everything other than "JMS"
- Minimal changes to make JMS pass tests
- rename endPoint -> url in FileParameter.
- Restore JMS from the master branch, and do the minimal changes such that we can "remove JmsReplyToDestination".

## Result

- `%message{key}` is still usable for FileDataInputParams/Http/Aggregating consumers
-  `%messageObject{}` is only supported in JMS stuffs.

## Testing

Turn on JMS testing, and check the coverage
Configure a request-reply style workflow such that 
- workflow 1 (doing the request is any XXX -> JMS)
- workflow 2 (doing the processing + reply) has a jms-producer that uses `%messageObject{JMSReplyTo}` as the endpoint.


aaron-mcgrath-adp:
I've rolled the src and tests back from the primary branch, made the minimum necessary changes to remove destinations.
Then made the minimum changes necessary to get the tests running.

One of the things we no longer support is the JmsReplyToDestination.   

Producers now support resolveable endpoints.
This is completely untested including the edge cases.

aaron-mcgrath-adp (approved):

PR #676, opened 2021-04-21
feature/interlok-3588 - FtpConsumer - Add the posibility to consume files from subfolders

albinoloverats:
## Motivation

The current FTP consumer will only consume files that are within the current working directory, without descending into any child directories. In certain situations it may be beneficial to scan sub directories too.

## Modification

Add a new consumer that will recursively descend into any sub directories it find.

## Result

All files found in sub directories get consumed.

## Testing

The unit tests mock a fake FTP server (which is how i discovered not all FTP servers are created equal or implement all commands). I also started an FTP server locally to confirm [adapter.xml](https://github.com/adaptris/interlok/files/6351491/adapter.txt).

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/ftp/FtpConsumer.java
+++ b/interlok-core/src/main/java/com/adaptris/core/ftp/FtpConsumer.java
@@ -16,12 +16,6 @@
 
 package com.adaptris.core.ftp;
 
-import static com.adaptris.core.AdaptrisMessageFactory.defaultIfNull;
-import static com.adaptris.core.ftp.FtpHelper.FORWARD_SLASH;
-import static org.apache.commons.lang3.StringUtils.isEmpty;
-import java.io.File;
-import java.io.FileFilter;
-import javax.validation.constraints.NotNull;
 import com.adaptris.annotation.AdapterComponent;
 import com.adaptris.annotation.AdvancedConfig;
 import com.adaptris.annotation.AutoPopulated;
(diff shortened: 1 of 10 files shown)


quotidian-ennui:
Should we make the metadata that's present in TraversingFsConsumer also shows up in the RecursiveFtpConsumer? 
It's only the addition of `CoreConstants.FS_CONSUME_PARENT_DIR` with "name" (not the full path) of the directory. If that's too hard to do w/o changing the underlying ftp client again, then it's fine to miss it off.

albinoloverats:
I've added the name of the parent of the current directory to the metadata:

    0 = {MetadataElement@4120} "key [fsConsumeDir] value [/home/user/work/dir2]"
    1 = {MetadataElement@4121} "key [fsParentDir] value [work]"
    2 = {MetadataElement@4122} "key [originalname] value [file0]"

or were you thinking more of just `dir2`?

quotidian-ennui:
> I've added the name of the parent of the current directory to the metadata:
> 
> ```
> 0 = {MetadataElement@4120} "key [fsConsumeDir] value [/home/user/work/dir2]"
> 1 = {MetadataElement@4121} "key [fsParentDir] value [work]"
> 2 = {MetadataElement@4122} "key [originalname] value [file0]"
> ```
> 
> or were you thinking more of just `dir2`?

I was thinking `dir2`, because I think that's the behaviour of TraversingFsConsumer (you might have to run a test mind).

aaron-mcgrath-adp (changes requested):
Nice idea to try changing directories and if passes then we know it's a directory...

I think we should re-add the javadoc to each field in the FtpConsumer.
Also, the new recursive consumer mentiones ConsumeDestinations in the class Javadoc, which doesn't exist anymore.

aaron-mcgrath-adp (approved):

PR #677, opened 2021-04-22
feature/interlok-3756

albinoloverats:
## Motivation

Allow for the support of expressions so that to generate an Xpath dynamically the user does not need to:

* Add `FormattedMetadata`: `key`=`xpath` `value`=`some/xpath/%message{value}`
* Then do `MetadataXpathQuery` and refer to it

But can instead do:

* `some/xpath/%message{value}` directly within the `XpathMetadataQuery`

## Modification

* Change the `createXpathQuery` method to resolve the query string (if there's a message - some existing tests do not pass in a message)
* Lombok-ify member variables where possible
* Add unit test

## Result

The user can now use expressions in any class that extends `ConfiguredXpathQueryImpl`

## Testing

To test, use [adapter.xml](https://github.com/adaptris/interlok/files/6358860/adapter.xml.txt) and add a couple of variable sets with variable `book-number` and a sensible value. Every minute the same message will be generated and the logs should show that the correct title is extracted from the XML and put into metadata.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/metadata/xpath/ConfiguredXpathQuery.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/metadata/xpath/ConfiguredXpathQuery.java
@@ -16,12 +16,11 @@
 
 package com.adaptris.core.services.metadata.xpath;
 
-import org.w3c.dom.Document;
-
 import com.adaptris.annotation.DisplayOrder;
 import com.adaptris.core.MetadataElement;
 import com.adaptris.util.text.xml.XPath;
 import com.thoughtworks.xstream.annotations.XStreamAlias;
+import org.w3c.dom.Document;
 
 /**
  * {@linkplain XpathQuery} implementation that retuns a single text item from the configured xpath.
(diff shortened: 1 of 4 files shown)


sebastien-belin-adp:
@albinoloverats It looks good but there are a few nice to have changes:

**In XpathQueryImpl**

- allowEmptyResults is now using lombok which is great but the javadoc has been removed at the same time as the getter and setter. You can add the javadoc above the field and that will still work.
- import com.adaptris.annotation.InputFieldHint is not used.

**In ConfiguredXpathQueryImpl**

- private String xpathQuery could have the UI annotation that let users know the field support expressions: @InputFieldHint(expression = true)
- import lombok.NonNull and import lombok.Setter are not used.
- In createXpathQuery(AdaptrisMessage msg) there is a null check, but is it something that can happen? Except ConfiguredXpathQueryCase#testCreateXpath ?

albinoloverats:
Yes, as far as I can tell `ConfiguredXpathQueryCase#testCreateXpath`is the only time the null check is needed in `createXpathQuery(AdaptrisMessage msg)`. I wasn't sure whether to keep the null check in or change the test.

sebastien-belin-adp:
> Yes, as far as I can tell `ConfiguredXpathQueryCase#testCreateXpath`is the only time the null check is needed in `createXpathQuery(AdaptrisMessage msg)`. I wasn't sure whether to keep the null check in or change the test.

Well I guess if we're not sure the null check is ok.
Approved

sebastien-belin-adp (approved):

PR #693, opened 2021-05-24
INTERLOK-3326 Conditional Logging

aaron-mcgrath-adp:
## Motivation

Our logging for conditional services is either too verbose in places or just doesn't show anything useful, therefore we will review the logging for the conditions and fix up where appropriate.

## Modification

We have a new logging method "logCondition" in the condition abstract class, all children conditions will call this method when they evaluate.  Additionally we have a new flag for this additionalLogging.  The flag defaults to false in case your data is sensitive and shouldn't be printed to logs.
The Operators now have a toString method that represents what they're testing; for example the Equals operator will return "equals myValue".
Any Condition that requires an Operator will call it's Operators toString method when building up the logging string; this means our conditions will log the condition type (Metadata/Payload/Expression etc), the source of the data and the boolean result of the evaluation.

## PR Checklist

- [x] been self-reviewed.
- [x] Added supporting annotations (like XStreamAlias / ComponentProfile)
- [x] Tested new/updated components in the UI

## Result

Optional useful logging for all conditional services, examples below;
`ConditionMetadata: evaluating key1 equals Value1 : result false`
`ConditionExpression: 1 == 2 evaluated to : false`
`ConditionPayload: evaluating payload equals some value : result false`
`ConditionMetadata: evaluating key1 is not in [val, val2, val3] : result true`

## Testing

Configure a conditional service, make sure you set the additionalLogging flag to true and then give it some activity.
Alternatively use the config enclosed.

[FS-Conditionals.txt](https://github.com/adaptris/interlok/files/6532501/FS-Conditionals.txt)

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/conditional/conditions/ConditionAnd.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/conditional/conditions/ConditionAnd.java
@@ -56,6 +56,7 @@ public class ConditionAnd extends ConditionListImpl {
         break;
       }
     }
+    logCondition("{}: evaluated to : {}", getClass().getSimpleName(), returnValue);
     return returnValue;
   }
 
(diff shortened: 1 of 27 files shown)


higgyfella (approved):
I tested this using various conditions in conditional services using the UI config page component test modal.

The logging all showed as expected. Approving PR.

PR #701, opened 2021-06-08
INTERLOK-3713 Decode base64 data

albinoloverats:
## Motivation

Service Tester cannot handle Excel files with a file-payload-provider in a test-case. This is due to serializable message, et al. all only handling strings.

## Modification

The DefaultSerializableMessageTranslator will now perform an Base64 decode to byte[] iff ContentEncoding is the string "BASE64".

## PR Checklist

- [x] been self-reviewed.

## Result

This will allows any binary file to be used in the service tester when using a file payload.

## Testing

To test this, follow the instructions on the [original ticket](https://adaptris.atlassian.net/browse/INTERLOK-3713).

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/DefaultSerializableMessageTranslator.java
+++ b/interlok-core/src/main/java/com/adaptris/core/DefaultSerializableMessageTranslator.java
@@ -25,6 +25,7 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
+import com.adaptris.util.text.mime.MimeConstants;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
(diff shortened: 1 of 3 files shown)


interlok-core/src/main/java/com/adaptris/core/DefaultSerializableMessageTranslator.java
```diff
@@ -89,6 +90,13 @@ public AdaptrisMessage translate(SerializableMessage message) throws CoreExcepti
         adaptrisMessage = messageFactory.newMessage(message.getContent(), message.getContentEncoding(),
             convertMap(message.getMessageHeaders()));
       }
+
+      if (message.getMessageHeaders().containsKey("_interlokMessageSerialization")) {
+        if (message.getMessageHeaders().get("_interlokMessageSerialization").equals("BASE64")) {
+          adaptrisMessage.setPayload(Base64.getDecoder().decode(message.getContent()));
```
quotidian-ennui, line 96: Probably need to make a constant out of `_interlokMessageSerilization` into CoreConstants so that you can use it in ServiceTester *easily*. Otherwise too much magic text.

quotidian-ennui (changes requested):

interlok-core/src/main/java/com/adaptris/core/DefaultSerializableMessageTranslator.java
quotidian-ennui: I think the if statement could be simplified to be (Base64 should probably be another constant??)

```
if ("BASE64".equalsIgnoreCase(message.getMessageHeaders().get("_interlokMessageSerilization") {
.. do stuff.
}
```

quotidian-ennui (approved):

PR #721, opened 2021-06-30
INTERLOK-3330 Deprecate ProduceExceptionHandler

albinoloverats:
## Motivation

We have already deprecated the restart-produce-exception-handler because it causes more problems than it solves. Currently there are only 3 implementations; one of them being a NullProduceExceptionHandler. One of last two has been deprecated which leaves the channel-restart-produce-exception-handler.

## Modification

* Add annotations, with messages and removal version info
* Update unit test

## PR Checklist

- [x] been self-reviewed.
- [x] Added supporting annotations (like XStreamAlias / ComponentProfile)

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/ChannelRestartProduceExceptionHandler.java
+++ b/interlok-core/src/main/java/com/adaptris/core/ChannelRestartProduceExceptionHandler.java
@@ -16,15 +16,21 @@
 
 package com.adaptris.core;
 
+import com.adaptris.annotation.Removal;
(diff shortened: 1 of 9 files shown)


quotidian-ennui:
WorkflowImp still contains a `setProduceExceptionHandler(new NullProduceExceptionHandler()`. That means that *all configuration* will report deprecations... so you clearly haven't checked your changes in the UI.

- Make the member default to null, make it nullable etc.
- Add protection to workflowImp#handleProduceException() and where produceExceptionHandler is used etc.

aaron-mcgrath-adp (approved):

PR #737, opened 2021-07-26
feature/INTERLOK-3543 - Add payload services

albinoloverats:
## Motivation

To scratch an itch - Some things could be accomplished easier if we had some basic payload manipulation like:

* append-to-payload-service
* merge-payloads-service (multi-payload)

If necessary, these can be moved out of this project into a new "com.adaptris.labs:developer-itch".

## Modification

Add new components:

* Payload from template referenced by payload/data input services
* Multi-payload aggregator
* Combine service (uses aggregators)

## PR Checklist

- [x] been self-reviewed.
- [x] Added javadocs for most classes and all non-trivial methods
- [x] Added supporting annotations (like XStreamAlias / ComponentProfile)
- [x] Added unit tests or modified existing tests to cover new code paths
- [x] Tested new/updated components in the UI and at runtime in an Interlok instance
- [x] Checked that javadoc generation doesn't report errors
- [x] Checked the display of the component in the UI

## Result

There are several new components for combining message payloads or building payloads from a template.

## Testing

Find attached a [simple config](https://github.com/adaptris/interlok/files/6877308/adapter.xml.txt) that combines multiple files from a directory into a multi-payload message and then combines all of those payloads into a single payload.

PR #761, opened 2021-08-26
feature/INTERLOK3828 - Resolvable expressions for file data/metadata

albinoloverats:
## Motivation

After putting forward a PR for #737 it was pointed out that we're trying to move away from data input/output parameters and towards an extensible expression language `%something{something-else}`. A conclusion was the reached that there needed to be something to replace `FileDataInputParameter`.

## Modification

Here is a file data/metadata resolver, that accepts the following format: `%file{/path/to/file:%size}` where %size could also be `%data`, `%type`, `%permissions`, `%date_create`, `%date_modify`, `%date_access`. The path can also be further resolvable, for instance from metadata: `%file{%message{file}:%data}`.

## PR Checklist

- [x] been self-reviewed.

## Result

As developers we can start to deprecate the data input/output parameters and provide a replacement for end users that is more powerful and easier to use and configure.

## Testing

Find attached an [adapter-config.xml](https://github.com/adaptris/interlok/files/7120899/adapter.xml.txt) that consumes a file, and using the metadata keys from the consumer, then reads some metadata about the file (it's actually file.wip but you know...).


interlok-common/src/main/java/com/adaptris/interlok/resolver/FileResolver.java
```diff
@@ -0,0 +1,294 @@
...
+          if (value == null)
+          {
+            log.warn("{} resolved to null for {}", what, path);
+            value = "";
+          }
+        }
+        catch (IOException e)
+        {
+          log.error("Could not resolve {} on path {} ", what, path, e);
+        }
```
quotidian-ennui, line 136: My only question would be, what's our expected behaviour if the file doesn't exist or the file is not "readable" by us.  If the answer is "undefined behaviour, possibly an exception" then that's probably cool, it's just a documentation issue...

Rather than logging an error and giving us something "meaningless" as the resolution (i.e. if the text was something like `%file{/path/exists:%size}-%file{/path/does/not/exist:%size}` we might end up with `%file{/path/does/not/exist:%size}` as the resolved text which could be completely unusable and untrackable.

You're "better off" failing with a UnresolveableException in all cases?

aaron-mcgrath-adp (approved):

PR #776, opened 2021-09-16
INTERLOK-1270

aaron-mcgrath-adp:
## Motivation

If you use startAdapterQuietly =true in your bootstrap.properties then any failure will be reported as a stack trace.  Problem is if you don't set this flag, then any errors are not reported.

## Modification
Simply made sure the bootstrap.start() is wrapped in a try/catch and any error reported.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/management/CmdLineBootstrap.java
+++ b/interlok-core/src/main/java/com/adaptris/core/management/CmdLineBootstrap.java
@@ -171,23 +171,27 @@ abstract class CmdLineBootstrap {
         @Override
         public void run() {
           Thread.currentThread().setName(threadName);
-          try {
-            bootstrap.start();
-          }
-          catch (Throwable e) {
-            System.err.println("(Error) Adapter Startup failure :" + e.getMessage());
-            logException(e);
-          }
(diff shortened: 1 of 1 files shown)


sebastien-belin-adp (commented):
@aaron-mcgrath-adp So basically now the difference between startAdapterQuietly=true or false is that when it's true it starts in a thread?
If that's so, does the name startAdapterQuietly still make sense?

aaron-mcgrath-adp:
> @aaron-mcgrath-adp So basically now the difference between startAdapterQuietly=true or false is that when it's true it starts in a thread?
> If that's so, does the name startAdapterQuietly still make sense?

Probably not, but it's been there for years and is well known by that name - changing it now would be like a config break.

sebastien-belin-adp:
> > @aaron-mcgrath-adp So basically now the difference between startAdapterQuietly=true or false is that when it's true it starts in a thread?
> > If that's so, does the name startAdapterQuietly still make sense?
> 
> Probably not, but it's been there for years and is well known by that name - changing it now would be like a config break.

I've raised a new issue for renaming the property INTERLOK-3841

sebastien-belin-adp (approved):

Issue #782, opened 2021-09-24
interlok-boot.jar gives inconsistent results depending on platform


quotidian-ennui:
# Scenario

I want to protect certain management endpoints (i.e. workflow-rest-services / workflow-health-check)
- I include dependencies in my build.grade that include the required jaspi jars.
- I fix up jetty.xml + webdefault.xml to setup a realm etc.
- I'm running in docker so docker build FTW

[docker-jetty-pw.zip](https://github.com/adaptris/interlok/files/7226753/docker-jetty-pw.zip)

# Steps to reproduce

- `docker build –tag lewinc/docker-jetty-pw .` 
- `docker run --rm -it -p8080:8080 lewinc/docker-jetty-pw`

This throws a NullPointerException, so nothing is started

```
WARN  [EmbeddedJettyStart] [o.e.j.w.WebAppContext.doStart()] [{}] Failed startup of context o.e.j.w.We
bAppContext@4386a299{Welcome to Interlok,/,file:///tmp/jetty-0_0_0_0-8080-ROOT_war-_-any-1600673352149
6383744/webapp/,UNAVAILABLE}{/opt/interlok/webapps/ROOT.war}
java.lang.NullPointerException
        at org.eclipse.jetty.security.jaspi.JaspiAuthenticatorFactory.getAuthenticator(JaspiAuthentica
torFactory.java:103)
        at org.eclipse.jetty.security.SecurityHandler.doStart(SecurityHandler.java:394)
        at org.eclipse.jetty.security.ConstraintSecurityHandler.doStart(ConstraintSecurityHandler.java
:417)
```

However, on your local machine, we can of course recreate the installation dir by assembling it.

- `cd builder && ./gradlew assemble`
- `cd build/distribution && java -Dinterlok.bootstrap.debug=true -jar lib/interlok-boot.jar`
- This starts up and `curl -si -uadmin:admin http://localhost:8080/workflow-health-check` gives you something meaningful.

# Investigations

If you inspect the output of the two windows, the *big difference* is that the InterlokLauncher debug is massively diferent in the ordering… on Windows it’s “order alphabetically”, on docker (alpine image) it’s whatever you want sort order…

I believe that this is a bug in the sense that even though everything is “the same”, you’re getting inconsistent results…

- In *this very specific case* the problem is apparently javaee-api.jar; which is being loaded before geronimo-jasp-spec.jar ?
[...]

PR #783, opened 2021-09-27
Force interlok-boot to resolve files alphabetically for consistency

quotidian-ennui:
## Motivation

This is a PR that alleviates the symptom of #782 (that's the motivation).

## Modification

Makes the scenario described by #782 more predictable (it doesn't solve it) by making the jars that are added by the spring-boot-classloader are ordered alphabetically vis String.compareTo rather than leaving it to the order returned by .listFiles() which may not be.

Did not make the utlimate List<Archive> (that is used by spring-boot-loader) a Treeset, because making the Archives sorted might have additional ramifications we don't understand. So, for a given directory, _ensure that the list of jar files is sorted alphabetically only_; the other behaviour is left as-is.

## PR Checklist

- [x] been self-reviewed.

## Result

Running against the docker image defined in the issue you now get, which resolves the specific symptom.

```
+ exec java -Xms488m -Xmx488m -Dpolyglot.js.nashorn-compat=true -Dpolyglot.engine.WarnInterpreterOnly=false -Dinterlok.bootstrap.debug=true -jar lib/interlok-boot.jar bootstrap.properties
(InterlokLauncher) Nested archive paths: [config/, lib/]
(InterlokLauncher) Main-Class: com.adaptris.core.management.SimpleBootstrap
(InterlokLauncher) Args for Main-Class : [bootstrap.properties]
(InterlokLauncher) Added config
(InterlokLauncher) Added lib
(InterlokLauncher) Adding jars from /opt/interlok/lib
(InterlokLauncher) Adding HdrHistogram.jar
(InterlokLauncher) Adding HikariCP.jar
(InterlokLauncher) Adding Saxon-HE.jar
(InterlokLauncher) Adding accessors-smart.jar
...
(InterlokLauncher) Adding geronimo-j2ee-management_1.1_spec.jar
(InterlokLauncher) Adding geronimo-jaspi.jar
(InterlokLauncher) Adding geronimo-jaspic_1.0_spec.jar
(InterlokLauncher) Adding geronimo-jms_1.1_spec.jar
(InterlokLauncher) Adding geronimo-jms_2.0_spec.jar
(InterlokLauncher) Adding geronimo-jta_1.1_spec.jar
...
(InterlokLauncher) Adding javaee-api.jar
...
(InterlokLauncher) Adding zjsonpatch.jar
```

This is also true running on WSL2 (Ubuntu20.04) + Mingw 64 (Git-Bash). This means that we have predictable behaviour even if we haven't figured out how javaee-api is screwing with jaspi (I think it might be the javax.security.auth.Message stuff still).

## Testing

Run against the described steps to reproduce in #782

### DIFF
--- a/interlok-boot/src/main/java/com/adaptris/interlok/boot/InterlokLauncher.java
+++ b/interlok-boot/src/main/java/com/adaptris/interlok/boot/InterlokLauncher.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.List;
 import org.springframework.boot.loader.Launcher;
 import org.springframework.boot.loader.PropertiesLauncher;
@@ -290,6 +291,7 @@ public class InterlokLauncher extends Launcher {
     }
     debug("Adding jars from " + dir.getCanonicalPath());
     File[] files = dir.listFiles(JAR_FILTER);
+    Arrays.sort(files, new FileComparator());
     for (File jar : files) {
       debug("Adding ", jar.getName());
       addArchive(jar, jars);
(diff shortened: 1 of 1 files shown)


quotidian-ennui:
@higgyfella You should also back port this (cherry pick or whatever) and do a 4.2.1 release; it's a pretty annoying issue that will burn docker users.

sebastien-belin-adp (approved):

PR #791, opened 2021-10-04
INTERLOK-3842

aaron-mcgrath-adp:
## Motivation

If you configure your expiring-map-cache to have a default of say 25 minutes for item expiration, but then add items with a custom expiry, you end up with many items that can each have a different expiry.
This is fine and intended, however if you absolutely don't want any item to exist in the cache past a single configured time period then this isn't possible.

We want to add a configuration item that can stop any item overriding the default expiry period.

## Modification

Very simply the expiring map cache now has a new boolean allow-expiry-overrides.  This defaults to 'true' which matches the old behaviour.  But when set to false and you add a new item to the cache, any custom expiry period is ignored.

## PR Checklist

- [x] been self-reviewed.

## Result

Default behaviour stays the same.  But now we can make sure that each item in the cache has the same expiry period if we want.

## Testing
The enclosed config accepts files into the "messages-in" directory and then adds an item to the cache if it does not already exist.
The cache is set to have a default of 1 minute expiry, but the put attempts to override this with 10 minutes.  Modifying the config of the cache to not allow overrides will change the behaviour as discussed above.

[fs-cache.txt](https://github.com/adaptris/interlok/files/7277766/fs-cache.txt)

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/cache/ExpiringMapCache.java
+++ b/interlok-core/src/main/java/com/adaptris/core/cache/ExpiringMapCache.java
@@ -21,6 +21,8 @@ import java.util.List;
 import java.util.concurrent.TimeUnit;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
+
+import org.apache.commons.lang3.BooleanUtils;
 import org.apache.commons.lang3.ObjectUtils;
 import com.adaptris.annotation.AdvancedConfig;
 import com.adaptris.annotation.AutoPopulated;
@@ -65,7 +67,9 @@ public class ExpiringMapCache implements Cache {
   @AutoPopulated
   @NotNull
   private ExpiringMapCacheListener eventListener;
-
+  @InputFieldDefault(value = "true")
+  private Boolean allowExpirationOverrides;
+  
   private transient ExpiringMap<String, Object> cache;
 
   public ExpiringMapCache() {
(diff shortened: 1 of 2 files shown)


quotidian-ennui:
There is an side effect to this change which is if you setup the cache to have a policy of LAST_ACCESSED then if you allow-overrides=false... then items with a custom expiry may in fact live _longer than you intended_ because custom expiry is from CREATION_TIME.

- Setup a cache with 25 minutes + LAST_ACCESSED+ allow-overrides=false
- Add an item in the cache with an TTL of 10 minutes.
- Use the item in the cache after in now + 7min
- The item will still exist in the cache in now() + 15 minutes when it really shouldn't.

This isn't documented which will; lead to unpredictable behaviour. If you are not going to rely on the default configuration for something like `GetAndCacheOauthToken`. What it would mean in that situation is you attempt to use an OAUTH token that is "no longer valid" where it will intermittently fail for what is observably _no easily recreatable reason_...

Also note that Cache#put(String,Object,long) is a default method defaults to throwing an UnsupportedOperationException.

aaron-mcgrath-adp:
Not a good fix.
Closing as "won't do".

quotidian-ennui:
@aaron-mcgrath-adp  -> arguably what you can do is to have a flag "allow-custom-expiry" that defaults to "TRUE"... Then if it's false you can just do `throw UnsupportedOperationException("as configured, custom expiry is not allowed")`. 

This I think gives you some of the behaviour that you want, and since only `GetAndCacheOauthToken` is the only thing that makes use of a custom expiry; you'll get better predictability.

Issue #808, opened 2021-10-15
JettyServerManager only expects to manage a single Server instance.


quotidian-ennui:
## Summary

JettyServerManager is capable of managing multiple server instances since it has an `add(Server)` method. However, if you are doing so then it attempts to add all the workflows to all the servers which will lead to potentially unpredictable behaviour

The problem is probably a design decision, since I think JettyServerManager.addDeployment() actually adds 
Perhaps we need to "key" the servers when we add them, and by default only regstier against the "jetty" one.

This leads us to the conclusion that if you had multiple jetty-http-connections configured, then all workflows that used *any* http-connection would be added to _all http-connections_.

## Steps to re-create

We can prove this by using various management components that use `jetty-embedded-connection` under the covers so use a standard build.gradle but with jolokia + health-check, interlokVersion == 4.3-SNAPSHOT (with jetty-9.4.44)

```
  interlokRuntime ("com.adaptris:interlok-json:$interlokVersion") { changing=true }
  interlokRuntime ("com.adaptris:interlok-jolokia:$interlokVersion") { changing=true }
  interlokRuntime ("com.adaptris:interlok-apache-http:$interlokVersion") { changing=true }
  interlokRuntime ("com.adaptris:interlok-rest-health-check:$interlokVersion") { changing=true }
```

Then in your bootstrap.properties you have (`rest.health-check.path` is just because I don't like the default)
```
managementComponents=jetty:rest:health-check:jolokia
rest.health-check.path=/health/*
```

Startup Interlok; in this instance I'm using the custom system properties to get some additional debugging out : `-Dinterlok.jetty.debug=true -Dinterlok.jetty.throw.unavailable.on.startup=true` it's not required, since the behaviour is the same.

So, at this point you can test the health check

```sh
$ curl -si http://localhost:8080/health
HTTP/1.1 200 OK
Content-Type: application/json
Transfer-Encoding: chunked

{"adapters":[{"adapter-state":{"id":"MyInterlokInstance","state":"StartedState","channel-states":[{"channel-state":{"id":"default-channel","state":"StartedState"}}]}}]}
```

_but_ the /health/ endpoint is __also__ registered against the jolokia-server-component that is created behind the scenes.

```sh
$ curl -si http://localhost:8081/health
HTTP/1.1 200 OK
Date: Fri, 15 Oct 2021 08:42:07 GMT
Content-Type: application/json
Transfer-Encoding: chunked
Server: Jetty(9.4.44.v20210927)

[...]

PR #809, opened 2021-10-15
JettyServerManager: multiple jetty server support

quotidian-ennui:
## Motivation

This fixes https://github.com/adaptris/interlok/issues/808

## Modification

JettyServerManager is now backed by a Map of Server instances with keys. Previous operations that don't use keys are marked as deprecated but should still work as expected.

If you don't use a key, then the deployment operations / add servlet operations operate on all server instances available to JettyServerManager. If you pass in a key then only the corresponding server instance is impacted.

EmbeddedConnection has been updated to refer to the new `JettyServerComponent#SERVER_ID`

Deprecated ServerManager since this is a redundant interface; there's only ever been a single implementation (an aborted tomcat impl notwithstanding) and the  assumption around using the javax.servlet isn't appropriate now (c.f. Quarkus/Spring Boot/Undertow).
- WSMU.getServerManager() -> deprecated, and points to JettyServerManager.getInstance() instead.

## PR Checklist

- [x] been self-reviewed.

## Result

This is an API change, so no impact on the UI

## Testing

- Test as per https://github.com/adaptris/interlok/issues/808. A warning will be logged by interlok-jolokia when it executes addServer, however the scenario listed in the corresponding issue should not occur (i.e. http://localhost:8081/health should now be a 404).

```
DEBUG [JolokiaEmbeddedJettyStart] [c.a.c.m.j.JolokiaComponent.run()] [{}] Creating Jetty wrapper
TRACE [JolokiaEmbeddedJettyStart] [c.a.c.m.j.JolokiaComponent.build()] [{}] Create Server from Jolokia Properties
WARN  [JolokiaEmbeddedJettyStart] [c.a.c.m.w.JettyServerManager.addServer()] [{}] No Key provided; generated key is 39a2f328-8002-454e-9feb-2c3a9b97d0a0
```

```console
$ curl -si http://localhost:8080/health
HTTP/1.1 200 OK
Content-Type: application/json
Transfer-Encoding: chunked

{"adapters":[{"adapter-state":{"id":"MyInterlokInstance","state":"StartedState","channel-states":[{"channel-state":{"id":"default-channel","state":"StartedState"}}]}}]}
$ curl -si http://localhost:8081/health
HTTP/1.1 404 Not Found
Cache-Control: must-revalidate,no-cache,no-store
Content-Type: text/html;charset=iso-8859-1
Content-Length: 450
Server: Jetty(9.4.44.v20210927)
```

### DIFF
--- a/interlok-common/src/main/java/com/adaptris/core/management/webserver/JettyServerManager.java
+++ b/interlok-common/src/main/java/com/adaptris/core/management/webserver/JettyServerManager.java
@@ -16,6 +16,7 @@
 
 package com.adaptris.core.management.webserver;
 
+import com.adaptris.interlok.util.Args;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -28,6 +29,10 @@ import java.util.Set;
 import java.util.UUID;
 import javax.servlet.Servlet;
 import javax.servlet.ServletContext;
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.Synchronized;
+import lombok.extern.slf4j.Slf4j;
 import org.eclipse.jetty.security.Authenticator;
 import org.eclipse.jetty.security.Authenticator.AuthConfiguration;
 import org.eclipse.jetty.security.ConstraintSecurityHandler;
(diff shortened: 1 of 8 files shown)


sebastien-belin-adp (approved):
We probably need to update interlok-jolokia to use the new code now :).

PR #819, opened 2021-11-04
JETTY : Initialise handlers in init() method not lazily

quotidian-ennui:
## Motivation
Lazy initialisation of handler map isn't suitable in high load environments since there is a chance of thread contention vis a vis state.

## Modification

Add new initialiseHandlers() method which removes the if null detection so that the handlers are no longer lazily initialised on first request.

## PR Checklist

- [x] been self-reviewed.

## Result

No configuration change
Bug fix


sebastien-belin-adp (approved):
Perfect, thank you

PR #854, opened 2021-12-16
feature/INTERLOK-2824 - seed based password implementation

albinoloverats:
## Motivation

Similar to what exists in PbeCrypto but uses a file instead. As the algorithm used in PbeCrypto is severely deprecated, and it relies on the network address of the host, it's in need of a replacement.

## Modification

Add new password-based encryption option, that uses a file, defined as such `-Dpassword.seed=/path/to/file`; it also uses more modern algorithms: PBEWithHmacSHA256AndAES_128.

Additionally, as PbeCrypto was deprecated and scheduled for removal in version 4.0.0 it has now been removed.

## PR Checklist

- [x] been self-reviewed.

## Result

The deprecated PbeCrypto is gone, and a new seed-based password-based encryption option is available.

## Testing

[This adapter.xml](https://github.com/adaptris/interlok/files/7728599/adapter.xml.txt) will demo that the new seed-based password implementation can encode/decode as expected.

NB Remember to add `-Dpassword.seed=/path/to/file` to your start script.


quotidian-ennui:
System.err logging?
I'm also not convinced that `password.seed` is the right system property, I'd consider `interlok.password.seed`

interlok-core/src/main/java/com/adaptris/security/password/SeededAesPbeCrypto.java
```diff
@@ -0,0 +1,164 @@
...
+      }
+      try (FileInputStream in = new FileInputStream(seedFile))
+      {
+        byte[] b = new byte[1048576];
+        int z = 0;
+        while ((z = in.read(b)) > 0)
+        {
+          sha.update(b, 0, z);
+        }
+      }
```
quotidian-ennui, line 154: while relying on apache is a trifle foolish in the wake of log4shell ;) commons.io means we don't necessarily have to write our own loops.
- Is someone's seed file going to be massive, of course, if someone uses a 10GB seed file, this is a silly idea.

```
    try (FileInputStream in = new FileInputStream(f)) {
      sha.update(IOUtils.readFully(in, in.available()));
    }
```

sebastien-belin-adp (approved):

PR #869, opened 2022-01-07
INTERLOK-2505 Combine common code for AbstractJdbcSequenceNumber + SimpleSequenceNumber

albinoloverats:
## Motivation

Currently, there is a lot of unnecessarily duplicated code between these two services.

## Modification

Pull a lot of the sequence number settings into their own class, which is then a member of each of the affected services.

## PR Checklist

- [x] been self-reviewed.

## Result

There will be a config change, although it should be minimal, and the end result from using the services should be the same.

## Testing

TODO: I'm putting together a test config.


quotidian-ennui:
@albinoloverats  -> Please don't forget that if you break config, then that's bad. Removing overflow-behaviour isn't great, you probably want to do some kind of deprecation warning blah blah blah on it

albinoloverats:
@quotidian-ennui indeed - this could require even more thought before becoming more than a draft.

Multiple inheritance would have been useful here but alas. Maybe a functional interface would achieve the desired result.

quotidian-ennui:
> @quotidian-ennui indeed - this could require even more thought before becoming more than a draft.
> 
> Multiple inheritance would have been useful here but alas. Maybe a functional interface would achieve the desired result.

Perhaps then we should flip it on its head and  deprecate the existing services, and create a new `UnifiedSequenceNumberService` that has pluggable storage ?  I'm not sure this is possible, but it's worth sketching out with dummy methods.

```java
public class Sequencer implements Service {
  private Storage storage;
  private String metadataKey;  
}

public class JdbcStorage implements Storage {
  private AdaptrisConnection connection;
  private String selectStatement; // etc etc etc.
}

public class SimpleStorage implements Storage {
  private String filename;
}
```

PR #873, opened 2022-01-14
Force SimpleDateFormat where used to be strict

quotidian-ennui:
## Motivation

```java
SimpleDateFormat format = new SimpleDateFormat("yyyyMMdd");
Date d = format.parse("2021-08-18");
```

That string with that format will _parse_ but the date parsed won't be 18th August 2021. It'll be some time in 2020. This is because SimpleDateFormat is by default lenient and it has accepted the first 8 characters as the date.

## Modification

- Add a strictFormatter() static method to DateFormatUtil
- Rather than creating a new SimpleDateFormat for every attempt to parse, make it a static ThreadLocal list of formatters in DateFormatUtil
- Where appropriate force delegation to the new strictFormatter() method rather than new SimpleDateFormat(). 
    - This is generally where it's obvious that we're going to use the `parse` method.
- Add a new explicit test to show difference between lenient/strict in DateFormatUtilTest.

## PR Checklist

- [x] been self-reviewed.

## Result

No visible change to the user; 

No existing tests are broken, so there is no regression; however, at runtime people may see _more parsing errors_ if they are running into the situation described above. That's actually better because having wrong dates is probably not a good thing. We have probably been getting away with it because a lot of the time we're transforming *from ISO8601 to something else* and a lenient ISO8601 parser is going to fail if `T` doesn't exist.

## Testing

- Use "ReformatDateService" and reformat the date from `yyyyMMdd` to `yyyy-MM-dd`
- Test the service against the latest release (using 2021-08-18 as your input), and check the new formatted key.
- Test the service against this PR branch...

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/AdapterStateSummary.java
+++ b/interlok-core/src/main/java/com/adaptris/core/AdapterStateSummary.java
@@ -1,12 +1,12 @@
 /*
  * Copyright 2015 Adaptris Ltd.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  *     http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


quotidian-ennui:
@higgyfella  -> You might want to raise a marker bug for your change log...

higgyfella (approved):
looks okay

PR #877, opened 2022-01-20
Add support for human readable sizes when creating size filters

quotidian-ennui:
## Motivation

If you want a composite file filter you can have the string `SizeGTE=<some number in bytes>__@@__Regex=.*\.xml` but it's hard to figure out the number of bytes, who remembers that 1048760 is 10Megabytes?

It would be easier to have a human readable form of the bytes.

## Modification

Since SizeBasedFilter takes a String argument we can take a human readable form and automagically convert it into the correct number of bytes.

Based on https://stackoverflow.com/a/52565585 

- 10G is transformed into 10 gigabytes
- 10M is transformed into 10 megabytes etc.
- Only the first char of the specifier counts, so 10MB = 10M
- Uppercase for the unit specifier so 10m == 10Megabytes.
- If no specifier (or an invalid one) is included, then defaults to bytes
- Has supports up to Yottabytes (which is largely irrelevant) to be honest.
- No support for SI style.

## PR Checklist

- [x] been self-reviewed.

## Result

There is no change for the user, since existing file-filter-implementations will take the raw bytes value as it always did.

## Testing

- Filesystem consumer with a file-filter-imp == SizeGreaterThanOrEqual
- Use something like 1k as the filter-expression.
- Start it, and create some files that match the the boundaries in question.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/fs/CompositeFileFilter.java
+++ b/interlok-core/src/main/java/com/adaptris/core/fs/CompositeFileFilter.java
@@ -59,7 +59,7 @@ import org.slf4j.LoggerFactory;
  * <p>
  * The filter only accepts the file if all the underlying filters accept the file, thus this
  * represents an implicit AND condition. So for example we would use
- * <code>SizeGT=4096__@@__Perl=.*\.xml</code> to match all files ending with <code>.xml</code> whose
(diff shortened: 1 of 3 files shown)


higgyfella (approved):
all looks good

PR #905, opened 2022-03-01
Add an InterlokMgmtComponentFilter to avoid "scripting"

quotidian-ennui:
## Motivation

If you are using scripting to do things in log4j2, this is now broken with 2.17.2. This is a case where the log4j2 team don't understand the concept of not changing expectations...

- https://github.com/adaptris/interlok/pull/903
- https://github.com/adaptris/interlok-build-parent/pull/85
- https://github.com/adaptris/interlok-build-parent/pull/86

This log4j2 snippet _no longer works_ unless you add the explicit system property ` -Dlog4j2.Script.enableLanguages=javascript`

```xml
<Configuration>
  <Scripts>
    <Script name="IsMgmtComponent" language="javascript">
      <![CDATA[
        if (logEvent.getContextMap().containsKey('ManagementComponent')) {
          true;
        } else {
          false;
        }
      ]]>
      </Script>
  </Scripts>
.... 
  <Loggers>
    <Root level="TRACE">
      <ScriptFilter onMatch="DENY" onMisMatch="NEUTRAL">
        <ScriptRef ref="IsMgmtComponent"/>
      </ScriptFilter>
      <AppenderRef ref="Console"/>
    </Root>
  </Loggers>
</Configuration>  
```

To remove the need for the system property, create a log4j2 filter that does precisely the same thing, but explicitly in java code

## Modification

Since non-default packages are not scanned by default, we create a new `logging.log4j2` package so that that we can specify it easily in the Configuraiton Element. 

- Add InterlokMgmtComponentFilter to log4j2 package
- Move JmxLogAppender to log4j2 package
- Modify build.gradle such that jacoco doesn't report on some things (com.adaptris.logging.jmx.*); this is testable, but a lot of mocking is required.

## PR Checklist

- [x] been self-reviewed.

## Result

There's no visual change for the user in the UI, but they can now configure a filter that magically filters out events where `ManagementComponent` is available in the diagnostic context.

## Testing

- Configure a parent gradle project that includes `interlok-rest-health-check`
- Configure your bootstrap.properties to use it
```
managementComponents=jmx:jetty:health-check
rest.health-check.path=/health-check/*
```
- Use a standard log4j2 configuration, and start up interlok
- `curl -si http://localhost:8080/health-check/alive`  which will return `200 OK`
    - You will see logging on the console from MetadataHandler + WorkflowHealthCheck.
- Reconfigure Log4j2.xml with
> Note that the packages= is not required, since we have used the annotation processor correctly so a Log4jPlugins.dat is created correctly in the jar file.

````
<Configuration status="warn" monitorInterval="60" shutdownHook="disable" packages="com.adaptris.logging.log4j2">
...
    <Root level="TRACE">
      <InterlokMgmtComponentFilter onMatch="DENY" onMismatch="NEUTRAL"/>
      <AppenderRef ref="Console"/>
    </Root>
</Configuration>
````
- Restart Interlok, and do the same curl again
    - This should have no logging on the console from MetadataHandler


higgyfella (approved):
Looks okay; approved

PR #935, opened 2022-04-19
INTERLOK-3890 # comment Add two new components  to send a multipart m…

sebastien-belin-adp:
## Motivation

We need to be able to send a file as part of a multipart/form-data request.

## Modification

Add two new components  to send a multipart message with Standard HTTP Producer:

- A http-multipart-message-request-headers which is a request-header-provider taking multipart mime message headers and adding them to the http request headers.
- A multipart-message-stream-input-parameter which is a request-body taking the multipart mime message content (without the headers) and adding it to the http request body.

## PR Checklist

- [x] been self-reviewed.
- [x] Checked that new 3rd party dependencies are appropriately licensed

## Result

The user can create a multipart MIME message containing a file with a multipart-message-builder.
Then the standalone-http-producer configured with the new http-multipart-message-request-headers and multipart-message-stream-input-parameter can send this multipart message.

## Testing

With a build of this branch you can have a configuration like the following:

```xml

  <!-- Read a file and put the content in the payload -->
  <read-file-service>
    <unique-id>jolly-lamarr</unique-id>
    <file-path>file:/path/to/pdf/file.pdf</file-path>
    <content-type-metadata-key>file-content-type</content-type-metadata-key>
  </read-file-service>

  <!-- Add a new multipart-message-builder service which will build your multipart message, assuming that your PDF is your payload -->
  <multipart-message-builder>
    <unique-id>build-multipart</unique-id>
    <inline-mime-body-part-builder>
      <body class="byte-array-from-payload"/>
      <part-header-filter class="fixed-values-metadata-filter">
        <metadata>
          <key-value-pair>
            <key>Content-Disposition</key>
            <value>form-data; name=&quot;file&quot;</value>
          </key-value-pair>
        </metadata>
      </part-header-filter>
      <content-type>application/pdf</content-type>
      <content-id>file</content-id>
    </inline-mime-body-part-builder>
    <mime-content-sub-type>form-data</mime-content-sub-type>
  </multipart-message-builder>

  <!--In the standalone-requestor you will need to use the two new components -->
  <standalone-requestor>
    <unique-id>attach-pdf-invoice</unique-id>
    <connection class="null-connection">
      <unique-id>null-connection</unique-id>
    </connection>
    <producer class="standard-http-producer">
      <unique-id>connect-caaps-api</unique-id>
      <url>http://localhost:8080/api/files</url>
      <method-provider class="http-configured-request-method">
        <method>POST</method>
      </method-provider>
     <ignore-server-response-code>true</ignore-server-response-code>
      <content-type-provider class="http-null-content-type-provider"/>
      <response-header-handler class="http-response-headers-as-metadata"/>
      <request-header-provider class="http-composite-request-headers">
        <http-multipart-message-request-headers>
          <unfold>true</unfold>
        </http-multipart-message-request-headers>
        <http-configured-request-headers>
          <headers>
            <key-value-pair>
              <key>X-Client-Id</key>
              <value>Some-Client-Id</value>
            </key-value-pair>
          </headers>
        </http-configured-request-headers>
      </request-header-provider>
      <request-body class="multipart-message-stream-input-parameter"/>
      <authenticator class="http-no-authentication"/>
    </producer>
  </standalone-requestor>
```

Then you will need an endpoint to which you can upload a multipart/form-data file with the name file.
That can be an endpoint you already have, a webapp, springboot app or even interlok with something like this:

```xml
<channel>
  <consume-connection class="jetty-embedded-connection">
    <unique-id>jetty-connection</unique-id>
  </consume-connection>
  <produce-connection class="null-connection">
    <unique-id>nostalgic-brahmagupta</unique-id>
  </produce-connection>
  <workflow-list>
    <standard-workflow>
      <consumer class="jetty-message-consumer">
        <message-factory class="default-message-factory"/>
        <unique-id>/api/files</unique-id>
        <destination class="configured-consume-destination">
          <configured-thread-name>/api/files</configured-thread-name>
          <destination>/api/files</destination>
        </destination>
        <parameter-handler class="jetty-http-parameters-as-metadata"/>
        <header-handler class="jetty-http-headers-as-metadata"/>
      </consumer>
        <unique-id>agitated-austin</unique-id>
        <services>
          <mime-part-selector-service>
            <unique-id>get-file-from-payload</unique-id>
            <selector class="mime-select-by-header">
              <header-name>Content-Disposition</header-name>
              <header-value-reg-exp>form-data; name=&quot;file&quot;</header-value-reg-exp>
            </selector>
          </mime-part-selector-service>
          <jetty-response-service>
            <unique-id>jetty-response</unique-id>
            <http-status>200</http-status>
            <content-type>text/plain</content-type>
          </jetty-response-service>
        </services>
      </service-collection>
      <producer class="fs-producer">
        <unique-id>write-file</unique-id>
        <destination class="configured-destination">
          <destination>file:/path/to/where/you/want/to/save/file</destination>
        </destination>
        <fs-worker class="fs-nio-worker"/>
        <filename-creator class="formatted-filename-creator">
          <filename-format>pdf-file.pdf</filename-format>
        </filename-creator>
      </producer>
      <produce-exception-handler class="null-produce-exception-handler"/>
      <unique-id>http</unique-id>
    </standard-workflow>
  </workflow-list>
  <unique-id>http</unique-id>
</channel>
```
Then you can check that the pdf file has been saved correctly in your file system.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
+++ b/interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2015 Adaptris Ltd.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.adaptris.core.http.client.net;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.util.Enumeration;
+
+import javax.mail.Header;
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMessage;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.adaptris.annotation.AdapterComponent;
+import com.adaptris.annotation.ComponentProfile;
+import com.adaptris.core.AdaptrisMessage;
+import com.adaptris.core.common.MultiPartMessageStreamInputParameter;
+import com.adaptris.core.http.client.RequestHeaderProvider;
+import com.thoughtworks.xstream.annotations.XStreamAlias;
+
+/**
+ * Implementation of {@link RequestHeaderProvider} that applies multipart MIME message headers as headers to a {@link HttpURLConnection}. It
+ * is meant to be used in conjunction with {@link MultiPartMessageStreamInputParameter}
+ *
+ * @config http-multipart-message-request-headers
+ *
+ */
+@XStreamAlias("http-multipart-message-request-headers")
+@AdapterComponent
+@ComponentProfile(summary = "Use a multipart message headers and add them as headers to an http connection", tag = "mime,http")
+public class MultiPartMessageRequestHeaders extends RequestHeaders implements RequestHeaderProvider<HttpURLConnection> {
+  protected transient Logger log = LoggerFactory.getLogger(this.getClass());
+
+  public MultiPartMessageRequestHeaders() {
+  }
+
+  @Override
+  public HttpURLConnection addHeaders(AdaptrisMessage msg, HttpURLConnection target) {
+    try {
+      MimeMessage mimeMessage = new MimeMessage(null, msg.getInputStream());
+      Enumeration<Header> headers = mimeMessage.getAllHeaders();
+
+      while (headers.hasMoreElements()) {
+        Header header = headers.nextElement();
+        String value = unfold(header.getValue());
+        log.trace("Adding Request Property [{}: {}]", header.getName(), value);
+        target.addRequestProperty(header.getName(), value);
+      }
+
+    } catch (MessagingException | IOException expts) {
+      log.error("Invalid multipart MIME message. No request header will be added.", expts);
+    }
+    return target;
+  }
+
+}
--- a/interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
+++ b/interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
@@ -86,7 +86,7 @@ public class MultiPartMessageRequestHeaders extends RequestHeaders implements Re
       }
 
     } catch (MessagingException | IOException expts) {
-      log.error("Invalid multipart MIME message. No request header will be added.", expts);
+      throw new RuntimeException("Invalid multipart MIME message", expts);
     }
     return target;
   }
--- a/interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
+++ b/interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
@@ -24,15 +24,21 @@ import javax.mail.Header;
 import javax.mail.MessagingException;
 import javax.mail.internet.MimeMessage;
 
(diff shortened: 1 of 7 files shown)


interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
```diff
@@ -0,0 +1,98 @@
...
+        if (excludeMessageId() && MimeConstants.HEADER_MESSAGE_ID.equals(header.getName())) {
+          log.trace("Ignoring Request Property [{}: {}]", header.getName(), value);
+        } else {
+          log.trace("Adding Request Property [{}: {}]", header.getName(), value);
+          target.addRequestProperty(header.getName(), value);
+        }
+      }
+
+    } catch (MessagingException | IOException expts) {
+      log.error("Invalid multipart MIME message. No request header will be added.", expts);
```
quotidian-ennui, line 89: Warning instead, since you're not throwing exception, do we need an extra debug flag gives us the exception, or just log it as part of the warning anyway?

interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
```diff
@@ -0,0 +1,98 @@
...
+  @InputFieldDefault("true")
+  private Boolean excludeMessageId;
+
+  public MultiPartMessageRequestHeaders() {
+  }
+
+  @Override
+  public HttpURLConnection addHeaders(AdaptrisMessage msg, HttpURLConnection target) {
+    try {
+      MimeMessage mimeMessage = new MimeMessage(null, msg.getInputStream());
```
quotidian-ennui, line 73: Ok, so in the situation with multipart-message-builder I think the payload becomse something like

```
Mime-Version: 1.0
Other headers: 

----MimeBoundary
asdfasdf
```

What about if the payload (because it wasn't generated by multipart-message-builder, and was consued from email or whatever) doesn't include the headers.

```
---MimeBoundary
asdfasdf
asdfasdf
```

MimeHelper#mimeFaker() method. Not sure it's reusable here, but there's going to have to be stuff.

interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
quotidian-ennui: Actually, no, should probably be a throw new RuntimeException() because it might break the server's expectations about how the payload should look?

interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
quotidian-ennui: This might not explicitly be the use-case because it's multipart/form-data, but shouldn't we basically "throw a runtime exception" and fail early rather than logging the error, and then just doing the wrong thing?

interlok-core/src/test/java/com/adaptris/core/common/MultiPartMessageStreamInputParameterTest.java
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  @Test(expected = CoreException.class)
+  public void testExtractWithException() throws Exception {
+    MultiPartMessageStreamInputParameter p = new MultiPartMessageStreamInputParameter();
+    AdaptrisMessage msg = new MyDefectiveMessage();
+    try (InputStream in = p.extract(msg)) {
+    }
+  }
+
```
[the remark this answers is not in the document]
sebastien-belin-adp: It's a very valid point.

interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
sebastien-belin-adp: I guess throwing an error is the right thing to do, even more so because we will rich it only when there is I/O error and not a bad formatting.

interlok-core/src/main/java/com/adaptris/core/http/client/net/MultiPartMessageRequestHeaders.java
sebastien-belin-adp: I guess throwing an error is the right thing to do, even more so because we will rich it only when there is I/O error and not a bad formatting.

PR #1000, opened 2022-09-12
INTERLOK-3904 Resolve service added.

aaron-mcgrath-adp:
## Motivation

Interlok has never handled basic string manipulation very well.  It's also one of the most requested features. 

## Modification

I've added a service that will allow you to specify a string which may include resolvable expressions which will be calculated at runtime.

## PR Checklist

- [x] been self-reviewed.


sebastien-belin-adp (approved):
@aaron-mcgrath-adp, could do with a bit more javadocs :p but otherwise it works greatly

PR #1010, opened 2022-09-21
owasp: add exclusion for jettison / xstream

quotidian-ennui:
## Motivation

Jettison-1.2 & XStream-1.4.19 marked with HIGH vulns > 7.5

Jettison is required because we use XStream to round-trip JSON but 1.3+ is marked as incompatible with XStream.

XStream is the cornerstone of config, so we have to suppress their DoS CVEs

Ref: http://x-stream.github.io/json-tutorial.html
Ref: http://x-stream.github.io/download.html#optional-deps

## Modification

- Add exclusions in for CVE-2022-40149 + CVE-2022-40150 for jettison
- Add exclusions for CVE-2022-40151 -> CVE-2022-40156 for xstream

## PR Checklist

- [x] been self-reviewed.

## Result

- `gradle dependencyCheckAnalyze` now works
- `gradle dependencyCheckAnalyze` now works for build-parent.

## Testing

- Run `gradle dCAnalyze` without the exclusions
- Run `gradle dCAnalyze` with the exclusions

### DIFF
--- a/gradle/owasp-exclude.xml
+++ b/gradle/owasp-exclude.xml
@@ -92,4 +92,27 @@
      <packageUrl regex="true">^pkg:maven/org\.eclipse\.jetty/jetty\-io@.*$</packageUrl>
      <vulnerabilityName>CVE-2022-2191</vulnerabilityName>
   </suppress>
+  <suppress>
+   <notes><![CDATA[
+   file name: jettison-1.2.jar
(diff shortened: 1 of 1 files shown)


sebastien-belin-adp:
@quotidian-ennui Thanks for that, I was looking into doing the same.
I think we will eventually have to test the compatibility with jettison 1.5 or migrate to an Xstream alternative.

PR #1020, opened 2022-10-13
INTERLOK-3936 Add the concept of payload metadata in multi payload message

sebastien-belin-adp:
## Motivation

Make it easier to access metadata related to a payload when using multi payload message and ForEach service.
This is mainly done to make accessing payload attachment easier in O365MailConsumer but can also have other benefits.

## Modification

- Add the concept of payload metadata in multi payload message. They are not proper payload metadata as they are available like all other metadata but they are prefixed with _PAYLOAD_my-payload-id__ which make them easily accessible in the for each service.
- Add some utility methods in MultiPayloadAdaptrisMessageImp to create, get and remove "payload metadata".
- Update ForEach service to add the payload id as metadata _PAYLOAD_ID_.
- Update ForEach service to add all payload metadata to the iterated message without the _PAYLOAD_my-payload-id__ prefix.  e.g. if the multipayload message has a payload with id my-payload-id and a metadata _PAYLOAD_my-payload-id_my-key_, when using the foreach the metadata _my-key_ will be also be available making it easier for some use cases.

## PR Checklist

- [x] been self-reviewed.

## Result

When using the ForEach service it will be easier to access metadata linked to a payload (for instance "emailattachmentfilename" and "emailattachmentcontenttype" generated added by O365MailConsumer).

## Testing

Use a build with interlok-core from this branch.

Have a simple config with one channel and one workflow.
The workflow should have a consumer that create a multi payload message.
Add a payload with id 'my-payload-id', add a metadata PAYLOAD_my-payload-id_my-key=my-value.
Add a ForEach service with a log message service.
Make sure that the logged message as the metadata my-key=my-value

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessage.java
+++ b/interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessage.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Writer;
+import java.util.Map;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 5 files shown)


higgyfella (approved):
seen it working; all good.

PR #1116, opened 2023-04-13
Pooling workflows and Shared connections fun.

aaron-mcgrath-adp:
## Motivation

If you have a pooling workflow that has references to a shared connection, then upon init/start you can get into a situation where multiple threads (each being a worker for the pooling workflow) attempt to register themselves as listeners to the shared connection.
This list of listeners in the shared connection was not thread safe and could cause an infinite loop resulting in Interlok app[earling to hang.

## Modification

Simple changed the maps holding the consumers/producers and listeners to be thread safe versions.

## Result

Interlok no longer hangs on startup when using a pooling workflow that references shared connections.


sebastien-belin-adp (approved):
Love those set initialisation :)

PR #1140, opened 2023-05-03
updated to fix fatal errors that were present in eclipse

higgyfella:
DRAFT PR - build failing on git, but working locally

## Motivation

So, I had 3 fatal errors when I imported interlok into eclipse.

## Modification

All 3 issues were due to lomok I think, so code calling a getter rather than a isser

## PR Checklist

- [x] been self-reviewed.

## Result

no errors in eclipse

## Testing

run tests, review changes


higgyfella:
It;'s weird that the gradle check is failing, because in eclipse everything is working.
perhaps my eclipse (using 17) has a different version of lombok?

the code clearly has
  @Getter
  @Setter
  private volatile boolean acceptSuccessCallbacks;
 
but here on git, there is no isAcceptSuccessCallbacks
and on eclipse there is no getAcceptSuccessCallbacks

PR #1163, opened 2023-06-01
Interlok 521 add regex validation for offset field

jameswickham:
## Motivation

So the offset field used by the AddTimestampMetadataService can be validated against a regex pattern that makes sure it adheres to the ISO8601 standard. This also includes validation in the UI and when the service is initialised.

## Modification

AddTimestampMetadataService class
AddTimestampMetadataServiceTest class
OffsetTimestampGenerator class

## PR Checklist

- [x] been self-reviewed.
- [n/a] Added supporting annotations (like @XStreamAlias / @ComponentProfile)

## Result

In the UI if the offset pattern they enter is not valid they will get an exception message telling them when they try to save.
If they edited the adapter xml directly with an invalid offset pattern they adapter will fail to start with the same exception message they would see in the UI.
If a message is sent through the service and the offset pattern is invalid they will also get an exception thrown(however this should not be possible due to the above two points)

## Testing

Load up a copy of interlok with the latest interlok-core.jar and add the following service
`<add-timestamp-metadata-service>
                <unique-id>offset-should-work</unique-id>
                <metadata-key>timestamp</metadata-key>
                <timestamp-generator class="offset-timestamp-generator">
                  <offset>invalid</offset>
                </timestamp-generator>
                <date-format-builder>
                  <format>yyyy-MM-dd&apos;T&apos;HH:mm:ssZ</format>
                </date-format-builder>
              </add-timestamp-metadata-service>`

Interlok should not start and throw the relevant exception.
To test in the UI do the same and try add the service and you should not even be able to save your changes as an exception should be thrown.

 **Also tested that this change did not break the service if 'LastMessageTimestampGenerator' is used.**

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/metadata/timestamp/OffsetTimestampGenerator.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/metadata/timestamp/OffsetTimestampGenerator.java
@@ -31,6 +31,9 @@ import com.adaptris.core.ServiceException;
 import com.adaptris.core.services.metadata.AddTimestampMetadataService;
 import com.thoughtworks.xstream.annotations.XStreamAlias;
 
+import lombok.Getter;
+import lombok.Setter;
+
 /**
  * Timestamp Generator implementation that mimics the default behaviour
  * available in {@link AddTimestampMetadataService}.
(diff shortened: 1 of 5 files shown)


interlok-core/src/main/java/com/adaptris/core/services/metadata/timestamp/OffsetTimestampGenerator.java
```diff
@@ -29,34 +31,43 @@
...
  * 
  * 
  * @see AddTimestampMetadataService
  * @since 3.5.0
  */
+
 @XStreamAlias("offset-timestamp-generator")
 public class OffsetTimestampGenerator implements TimestampGenerator {
 
+  private static final String OFFSET_VALIDATION_REGEX = "^\\-?P(?=\\w*\\d)(?:\\d+Y|Y)?(?:\\d+M|M)?(?:\\d+D|D)?(?:T(?:\\d+H|H)?(?:\\d+M|M)?(?:\\d+(?:\\­.\\d{1,2})?S|S)?)?$";
```
sebastien-belin-adp, line 45: That's some pretty cool regex

interlok-core/src/main/java/com/adaptris/core/services/metadata/timestamp/OffsetTimestampGenerator.java
```diff
@@ -29,34 +31,43 @@
...
     setOffset(offset);
   }
 
-
   @Override
   public Date generateTimestamp(AdaptrisMessage msg) throws ServiceException {
     Date timestamp = new Date();
     try {
       if (!isBlank(offset)) {
+        java.util.regex.Pattern pattern = java.util.regex.Pattern.compile(OFFSET_VALIDATION_REGEX);
```
sebastien-belin-adp, line 66: Technically you don't need that, and I wonder if it may also add some slow down on the message processing as this is compiling the pattern and validating the offset for every message.
The annotation should be enough as, in the UI or when enabling validation at startup the offset will be validated.

PR #1167, opened 2023-06-08
Interlok 4071 expose timeout options

jameswickham:
## Motivation

So we can expose the timeout options that are available in the standard http producer.

## Modification

Updated the class HttpRequestServiceImp so it gives the timeout options in the UI and when creating a StandardHttpProducer object we pass in the timeout values that have been set.

Also added a new unit test that tests that the read timeout will throw an exception if it is exceeded.

## PR Checklist

- [x] been self-reviewed.

## Result

The end user is now able to choose and swet connect and read timeout durations when using this service.

## Testing

You can create a channel->workflow that creates a request using a polling consumer and the HttpRequestService(making sure you have the read timeout set) then create a second channel->workflow that can mock a http endpoint by using a Jetty connection and consumers. Within that workflow include a wait service followed by a jetty response service. Make sure the wait service time period is longer than whatever the read timeout is.


higgyfella (approved):
looks good

Issue #866, opened 2022-01-04
JDBC Stored procedure tests fail with SQL Server


quotidian-ennui:
# Expectation

If you switch to MSSQL as opposed to MySQL to run the stored procedure tests, then the tests should pass in the same way, they don't against the standard microsoft sql server image. Most of the tests do work, enough don't for it to be investigated.

Note that I tried this with _9.2.1.jre8_ (current develop branch) and _9.4.1.jre11_ (jre11 since we're java 11 for interlok 4.x)

# Steps to recreate

### MSSQL in docker

The docker compose file is this (very simple).

```yaml
version: '3.2'
services:
  mssql_server:
    image: mcr.microsoft.com/mssql/server:2019-latest
    hostname: mssql_server
    container_name: mssql_server
    environment:
      - SA_PASSWORD=Yukon900
      - ACCEPT_EULA=Y
    ports:
      - "1433:1433"
```
### properties configuration 

I used `src/test/resources/default-test.properties.template.[machine-name]` to override the defaults.

```properties
default.jdbc.storedproc.tests.enabled=true
default.jdbc.storedproc.url=jdbc:sqlserver://localhost:1433;databaseName=
default.jdbc.storedproc.driver=com.microsoft.sqlserver.jdbc.SQLServerDriver
default.jdbc.storedproc.database=InterlokTest
default.jdbc.storedproc.username=sa
default.jdbc.storedproc.password=Yukon900
default.jdbc.storedproc.teststatement=
# Currently supported vendors for testing; "mysql" / "sqlserver".
default.jdbc.storedproc.vendor=sqlserver
default.jdbc.storedproc.url.params=
```

### Test Output

```console
Running: Test class com.adaptris.core.jdbc.StoredProcedureProducerTest

com.adaptris.core.jdbc.StoredProcedureProducerTest > testMultiResultSetXmlTranslator FAILED
    java.lang.AssertionError at StoredProcedureProducerTest.java:1013

com.adaptris.core.jdbc.StoredProcedureProducerTest > testMultipleResultSetsFirstRowMetadataTranslator FAILED
    java.lang.AssertionError at StoredProcedureProducerTest.java:954

com.adaptris.core.jdbc.StoredProcedureProducerTest > testMultipleResultSetsMetadataTranslator FAILED
    java.lang.AssertionError at StoredProcedureProducerTest.java:910
Running: Test class com.adaptris.jdbc.JdbcStoredProcedureTest

[...]

PR #1189, opened 2023-07-04
Adding a condition to check which JDBC vendor is set and only run the…

jameswickham:
## Motivation

fixes https://github.com/adaptris/interlok/issues/866

This is due to the fact that the 'sqlserver' vendor does not support multiple open result sets. Page can be found here: https://learn.microsoft.com/en-us/sql/connect/jdbc/reference/supportsmultipleopenresults-method-sqlserverdatabasemetadata?view=sql-server-ver16

## Modification

Added conditions to the two unit tests so they will check which vendor is being used and run the tests accordingly.

## PR Checklist

- [x] been self-reviewed.
- [n/a] Added validation annotation (@NotNull...) when required and add @Valid when nested class has some validation
- [n/a] Added comments explaining the "why" and the intent of the code wherever it would not be obvious for an unfamiliar reader
- [n/a] Tested new/updated components in the UI and at runtime in an Interlok instance
- [n/a] Checked the display of the component in the UI

## Result

No end result for the user, the change is only to how the unit tests are run.

## Testing

Run the two modified unit tests:
JdbcStoredProcedureTest.java
StoredProcedureProducerTest.java

try with both the 'mysql' and then the 'sqlserver' vendors and confirm all the unit tests still run. How to configure the vendors for running these tests is defined in this repo's readme and also some information in the issue: https://github.com/adaptris/interlok/issues/866

### DIFF
--- a/interlok-core/src/test/java/com/adaptris/core/jdbc/StoredProcedureProducerTest.java
+++ b/interlok-core/src/test/java/com/adaptris/core/jdbc/StoredProcedureProducerTest.java
@@ -57,6 +57,9 @@ import com.adaptris.util.TimeInterval;
 import com.adaptris.util.XmlUtils;
 import com.adaptris.util.text.xml.XPath;
 
+import lombok.Getter;
+import lombok.Setter;
+
 public class StoredProcedureProducerTest extends ExampleProducerCase {
 
   /**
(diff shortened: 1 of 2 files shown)


quotidian-ennui:
fixes https://github.com/adaptris/interlok/issues/866

aaron-mcgrath-adp (approved):
Nice.

PR #1232, opened 2023-08-22
Removing the use of the internal security manager.

aaron-mcgrath-adp:
## Motivation

The internal SecurityManager is deprecated for future removal.

## Modification

We were using the SecurityManager to gain access to a ThreadGroup; however the default behaviour appears to return the current threads group, which doesn't need the SecurityManager at all.  So we have just removed the line to that class and get access to the current threads group manually.

## Result

No warnings about SecurityManager being deprecated.

## Testing

nothing changes.


sebastien-belin-adp (approved):
That's exactly what I would have done.

PR #1249, opened 2023-09-13
Feature/interlok 4123 payload path encryption

jameswickham:
## Motivation

Adding two new Interlok services and supporting classes that will allow you to encrypt/decrypt a payload with a pointable path.

## Modification

Two new service classes have been added: PayloadPathEncryptionService & PayloadPathDecryptionService - these classes extend the CoreSecurityService abstract class so they have enherited all of the actual encryption/decryption functionality from this.
A new Interface: PathBuilder 
A class that implements the Interface: XpathBuilder
Test classes for the above that have been added.
A small update to the XMLHelper class, a new method to convert a string to a node.

## PR Checklist

- [x] been self-reviewed.

## Result

Two new services that you can define paths for that will encrypt/decrypt the content of those paths in a given message.

## Testing

To test this you can unzip the file which is a v5 adapter installation with all the required files: https://cloud.adaptris.net/index.php/s/goVlkU0fJJKWg8P

So do the following:

1. unzip and make sure your %JAVA_HOME% is pointing to a Java 17 installation.
2. loads up the project 'PathSecurityTesting' which is configured to fire off 4 messages every second(2 json, 2 xml) each of those will have a message that also runs through a jolt or xslt mapping.
4. If your home path is different to 'C:\Adaptris\' then update the keystore URLS to your home path.
5. You can also run the service tester in the UI or using the command: java -jar lib\interlok-boot.jar -serviceTest src\test\interlok\service-test.xml

The service tests should pass with the asseration that the payload matches the input(except in the two cases where we are transforming the data and then decrypting which will then match the expected mapped payload with decrypted data)

Even with 4 messages going through a second the adapter seems to be running ok and not struggling or taking up too much memory.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/security/XpathBuilder.java
+++ b/interlok-core/src/main/java/com/adaptris/core/security/XpathBuilder.java
@@ -22,6 +22,7 @@ import com.adaptris.annotation.AdvancedConfig;
 import com.adaptris.annotation.AutoPopulated;
 import com.adaptris.annotation.ComponentProfile;
 import com.adaptris.annotation.DisplayOrder;
+import com.adaptris.annotation.InputFieldHint;
 import com.adaptris.core.AdaptrisMessage;
 import com.adaptris.core.CoreException;
 import com.adaptris.core.ServiceException;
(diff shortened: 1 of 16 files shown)


sebastien-belin-adp (approved):
@jameswickham 
I'm approving this PR. Good work.
There were a few missing tests that I've added and a few places that I did a bit of refactoring. (Like avoid to use continue when an 'else' can do the job).
I renamed xpaths to xpath in the xml path-builder. So that will break your existing config.
Tell me if I'm wrong but I think they are entities part of the list so it's better to not have the 's'.
We should do the same for the json-path I guess.

interlok-core/src/main/java/com/adaptris/core/security/XpathBuilder.java
```diff
@@ -134,20 +144,16 @@ public void insert(AdaptrisMessage msg, Map<String, String> pathKeyValuePairs) t
...
-            try {
-              Node nestedNodes = XmlHelper.stringToNode(xpathValue);
-              Node nestedNodesToImport = doc.importNode(nestedNodes, true);
-              while (node.hasChildNodes()) {
-                node.removeChild(node.getFirstChild());
-              }
-              node.appendChild(nestedNodesToImport);
-            } catch (Exception e) {
-              throw new ServiceException(NON_XML_EXCEPTION_MESSAGE);
+        if (xpathValue.startsWith("<") && isStringNestedNodes(xpathValue)) {
```
jameswickham, line 148: Just wanted to add the only reason I had these as anotehr nested IF statement was I was thinking that it would only be worth checking messages against the isStringNestedNodes() method only if they start with < as to not run that check on all messages. I didn't stop and think that if the firt condition returns false it wouldn't bother checking the 2nd condition as it's linked with &&.

PR #1431, opened 2024-02-12
Bug/interlok 4325 - MultiPayloadAdaptrisMessage issue where original payload cannot be updated if content originated from an input stream.

jameswickham:
## Motivation

To Address a bug where when a multipayload message is created and we call the method
`public OutputStream getOutputStream(@NotNull String payloadId)` this was causing the Payload class to store a ByteArrayOutputStream against the variable 'stream' which meant that when you tried to update the payloads content logic in the payload() method was always setting the payload to the value originally stored against the variable 'stream'
## Modification

Added a new unit test and supproting files to:
A. prove the bug existed(as the unit test was failing as expected)
B. to show the bug is now fixed as the unit test now passes.

I also implemented a small fix within the inner class Payload so we no longer store the two different data types and then have logic to decide which one to retrive as when the data is retrieved it's always as a byte array so I altered the logic so we always store the data as a byte array. The way the class is used by the calling code has not changed.

## PR Checklist

- [x] been self-reviewed.
- [n/a ] Added supporting annotations (like @XStreamAlias / @ComponentProfile)
- [n/a ] Added DefaultValue annotation when there is a default value (e.g. @DefaultValue('true'))
- [n/a] Added validation annotation (@NotNull...) when required and add @Valid when nested class has some validation
- [n/a] Tested new/updated components in the UI and at runtime in an Interlok instance
- [] Checked that javadoc generation doesn't report errors - **_does report errors but these have been fixed in a PR that is yet to be merged(https://github.com/adaptris/interlok/pull/1254)_**
- [n/a] Checked the display of the component in the UI

## Result

they should not notice anything different.

## Testing

Run unit test without the code change and it will fail and the result shows the payload is not being updated from wha it originally was. After the code change is in place run the unit test again and you see it passes and what the test asserts is that the payload is updated and matches the new expected payload.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessageImp.java
+++ b/interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessageImp.java
@@ -622,11 +622,13 @@ public class MultiPayloadAdaptrisMessageImp extends AdaptrisMessageImp implement
 
   private class ByteFilterStream extends FilterOutputStream {
     private final String payloadId;
+    private Payload pl;
 
     ByteFilterStream(@NotNull String payloadId, ByteArrayOutputStream out) {
       super(out);
       this.payloadId = payloadId;
-      payloads.put(payloadId, new Payload(out));
+      this.pl = new Payload(out);
+      payloads.put(payloadId, pl);
     }
 
     @Override
@@ -639,7 +641,6 @@ public class MultiPayloadAdaptrisMessageImp extends AdaptrisMessageImp implement
   private class Payload {
     String encoding = getFactory().getDefaultCharEncoding();
     private byte[] data;
-    private ByteArrayOutputStream stream;
 
     Payload(String encoding, @NotNull byte[] data) {
       this.encoding = encoding;
@@ -651,13 +652,10 @@ public class MultiPayloadAdaptrisMessageImp extends AdaptrisMessageImp implement
     }
 
     Payload(@NotNull ByteArrayOutputStream stream) {
-      this.stream = stream;
+      this.data = stream.toByteArray();
     }
 
     byte[] payload() {
-      if (stream != null) {
-        return stream.toByteArray();
-      }
       return data;
     }
   }
--- a/interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessageImp.java
+++ b/interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessageImp.java
@@ -626,7 +626,7 @@ public class MultiPayloadAdaptrisMessageImp extends AdaptrisMessageImp implement
     ByteFilterStream(@NotNull String payloadId, ByteArrayOutputStream out) {
       super(out);
       this.payloadId = payloadId;
-      payloads.put(payloadId, new Payload(out));
+      addPayload(payloadId, ((ByteArrayOutputStream) super.out).toByteArray());
     }
 
     @Override
@@ -647,10 +647,6 @@ public class MultiPayloadAdaptrisMessageImp extends AdaptrisMessageImp implement
 
     Payload(@NotNull byte[] data) {
       this.data = data;
-    }
-
-    Payload(@NotNull ByteArrayOutputStream stream) {
-      this.data = stream.toByteArray();
     }
 
     byte[] payload() {
--- a/interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessageImp.java
+++ b/interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessageImp.java
@@ -622,13 +622,11 @@ public class MultiPayloadAdaptrisMessageImp extends AdaptrisMessageImp implement
 
   private class ByteFilterStream extends FilterOutputStream {
     private final String payloadId;
-    private Payload pl;
 
     ByteFilterStream(@NotNull String payloadId, ByteArrayOutputStream out) {
       super(out);
       this.payloadId = payloadId;
-      this.pl = new Payload(out);
-      payloads.put(payloadId, pl);
+      payloads.put(payloadId, new Payload(out));
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessageImp.java
```diff
@@ -651,13 +650,10 @@ private class Payload {
     }
 
     Payload(@NotNull ByteArrayOutputStream stream) {
-      this.stream = stream;
+      this.data = stream.toByteArray();
```
sebastien-belin-adp, line 653: I think, if you go this way, you may as well change line 629 ` payloads.put(payloadId, new Payload(out));`
by something like line 635 `addPayload(payloadId, ((ByteArrayOutputStream) super.out).toByteArray());`

Then you can remove this constructor altogether as it was only used line 629.

I'm not really sure why it was doing that but it may worth doing more tests with `MultiPayloadStreamOutputParameter` and `MultiPayloadXmlTransformService` as they are using `msg.getOutputStream(outputPayloadId)` which eventually call the code we're talking about.

In fairness, I don't really know why this code was like that and if we are not breaking something removing it. But I'd otherwise agree with your point.

interlok-core/src/test/java/com/adaptris/core/MultiPayloadAdaptrisMessageImpTest.java
```diff
@@ -62,5 +67,22 @@ public void testRemovePayloadMessageHeader() {
...
+    System.out.println(multiPayloadAdaptrisMessage.getContent("default-payload"));
+    multiPayloadAdaptrisMessage.setContent("new payload", multiPayloadAdaptrisMessage.getContentEncoding());
+    
+
+    assertEquals("new payload", multiPayloadAdaptrisMessage.getContent());
+    
+    multiPayloadAdaptrisMessage.addContent("default-payload", "another payload", multiPayloadAdaptrisMessage.getContentEncoding());
+
+    assertEquals("another payload", multiPayloadAdaptrisMessage.getContent());
+
```
sebastien-belin-adp, line 85: Remove empty line

interlok-core/src/test/java/com/adaptris/core/MultiPayloadAdaptrisMessageImpTest.java
```diff
@@ -62,5 +67,22 @@ public void testRemovePayloadMessageHeader() {
...
   }
+  
+  //Unit test added to test bug INTERLOK-4325 
+  @Test
+  public void testUpdatingOriginalPayloadFromInputStream() throws IOException {
+	System.out.println(PROPERTIES.getProperty(EXAMPLE_PAYLOAD_FILE));
+	MultiPayloadAdaptrisMessage multiPayloadAdaptrisMessage = MessageHelper.createMultiPayloadMessage("default-payload", PROPERTIES.getProperty(EXAMPLE_PAYLOAD_FILE));
+    System.out.println(multiPayloadAdaptrisMessage.getContent("default-payload"));
+    multiPayloadAdaptrisMessage.setContent("new payload", multiPayloadAdaptrisMessage.getContentEncoding());
+
```
sebastien-belin-adp, line 78: Remove one empty line.

interlok-core/src/test/java/com/adaptris/core/MultiPayloadAdaptrisMessageImpTest.java
```diff
@@ -62,5 +67,22 @@ public void testRemovePayloadMessageHeader() {
     assertFalse(multiPayloadAdaptrisMessage.payloadHeadersContainsKey("key"));
     assertFalse(multiPayloadAdaptrisMessage.payloadHeadersContainsKey("another-payload", "key"));
   }
+  
+  //Unit test added to test bug INTERLOK-4325 
+  @Test
+  public void testUpdatingOriginalPayloadFromInputStream() throws IOException {
+	System.out.println(PROPERTIES.getProperty(EXAMPLE_PAYLOAD_FILE));
+	MultiPayloadAdaptrisMessage multiPayloadAdaptrisMessage = MessageHelper.createMultiPayloadMessage("default-payload", PROPERTIES.getProperty(EXAMPLE_PAYLOAD_FILE));
+    System.out.println(multiPayloadAdaptrisMessage.getContent("default-payload"));
```
sebastien-belin-adp, line 76: We want to avoid System.out.println.

jameswickham: gahhh ailly mistake and forgot ot remove this, removed now!

interlok-core/src/test/java/com/adaptris/core/MultiPayloadAdaptrisMessageImpTest.java
jameswickham: removed unneeded empty lines and just left the 1 empty line between the two assert equals for ease of reading(but can remove this line if better)

interlok-core/src/test/java/com/adaptris/core/MultiPayloadAdaptrisMessageImpTest.java
jameswickham: removed unneeded empty lines and just left the 1 empty line between the two assert equals for ease of reading(but can remove this line if better)

interlok-core/src/main/java/com/adaptris/core/MultiPayloadAdaptrisMessageImp.java
jameswickham: I reviewed the classes you mentioned and it looks like doni what you suggested doesn't break these.

I had a look at what the test classes for these are doing and in the tests I can see they are taking an input stream, writing it to the messages output stream and then adding it as a new payload.

After making the changes these tests are still working as expected so is it safe to say that it is worknig as expected as it still successfully writes to the messages output stream AND is able to add it as a new payload to the multi payload message, it can also retrive the payload to check it's content.

sebastien-belin-adp (approved):
Good work

PR #1480, opened 2024-04-02
Added a timeout for the URLConnection in the URLHelper

aaron-mcgrath-adp:
## Motivation

Any service using the URLHelper to fetch a remote resource could spin forever, never timing out.
This would cause a hang.

## Modification

Simply added the read and connect timeouts defaulted to a sensible 30 seconds to the URLConnection  code in the URLHelper.


sebastien-belin-adp (approved):
Should we make it configurable? But I guess that will do for now.

PR #1684, opened 2024-10-30
INTERLOK-4445 Fix multi payload message mime encoder working with the encoding and decoding service.

sebastien-belin-adp:
## Motivation

When using a multipayload message there is an issue with MultiPayloadAdaptrisMessageImp#ByteFilterStream that resets the default payload when doing msg.getOutpuStream

## Modification

Use a ByteArrayOutputStream first and then set the byte payload

## PR Checklist

- [x] been self-reviewed.

## Result

The user can use EncodingService and DecodingService with multi payload message

## Testing

```xml
<encoding-service>
  <unique-id>to-mime-message</unique-id>
  <encoder class="com.adaptris.core.MultiPayloadMessageMimeEncoder"/>
</encoding-service>

<decoding-service>
  <unique-id>from-mime-message</unique-id>
  <encoder class="com.adaptris.core.MultiPayloadMessageMimeEncoder"/>
  <message-factory class="multi-payload-message-factory">
    <default-char-encoding>UTF-8</default-char-encoding>
    <default-payload-id>default-payload</default-payload-id>
  </message-factory>
</decoding-service>
```
Use a multi payload message with payload:
id: 'default-payload'
```
<xml>content</xml>
```
And metadata
```
key=value
```

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/MultiPayloadMessageMimeEncoder.java
+++ b/interlok-core/src/main/java/com/adaptris/core/MultiPayloadMessageMimeEncoder.java
@@ -84,6 +84,7 @@ public class MultiPayloadMessageMimeEncoder extends MimeEncoderImpl<OutputStream
       }
       writeNextServiceId(output, msg);
       output.writeTo(target);
+      target.flush();
     } catch (Exception e) {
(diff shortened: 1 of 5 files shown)


higgyfella (approved):
looks good

PR #1707, opened 2024-11-15
INTERLOK-4430 When using a URL to configure a xml transform service w…

sebastien-belin-adp:
This should allow us to keep the relative path import working.

## Motivation

We did some work 6-12 months ago where we enabled InputParameters rather than a simple input string as the source for your XSL.

This essentially has changed the xml transformer creation in that we don’t give the URL of the XSL to the transformer factory, but instead an input stream of the file XSL content.  The unintentional consequence of this is that now the transformer no longer has file location context.  Which means, that if you have an import statement in your XSL that is relative to the source XSL the transformer cannot resolve it.

## Modification

When using a URL to configure a xml transform service we directly create a transformer using the url rather than the content of the xsl retrived from the url. This should allow us to keep the relative path import working.

Added test to make sure that's working0

Remove deprecation of url as this is the only way of using xsl:import and caching.

## PR Checklist

- [x] been self-reviewed.

## Result

The user should be able to use the XML transform service as before the changes made for the migration tool when using a URL.
Meaning that if they us an import with relative path, the relative path will be calculated from the url.
If using a string content for the XML transform service it should work the same (not working with import).

## Testing

With the interlok-core.jar from this PR

Have two xslt in the same dir:

simple-transform-xsl-import.xsl:
```xml
<?xml version='1.0' encoding='utf-8' ?>
<xsl:stylesheet 
  xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0" 
  xmlns:java="http://xml.apache.org/xslt/java" exclude-result-prefixes="java">
  
  <xsl:param name="world"/>
  
  <xsl:output method="text"/>
  
  <xsl:template match="/">  
    <xsl:value-of select="//payload"/>
    <xsl:value-of select="$world"/>
  </xsl:template>
   
</xsl:stylesheet>
```
and

simple-transform.xsl
```xml
<?xml version='1.0' encoding='utf-8' ?>
<xsl:stylesheet 
  xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0" 
  xmlns:java="http://xml.apache.org/xslt/java" exclude-result-prefixes="java">
  
  <xsl:import href="./simple-transform.xsl"/>
   
</xsl:stylesheet>
```

A message with payload:

```xml
<message>
  <payload>Hello</payload>
</message>
```
And service XML
```xml
<xml-transform-service>
  <unique-id>prickly-curie</unique-id>
  <mapping-source class="file-data-input-parameter">
    <url>file:///paht/to/simple-transform-xsl-import.xsl</url>
  </mapping-source>
</xml-transform-service>
```

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/transform/XmlTransformService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/transform/XmlTransformService.java
@@ -68,8 +68,11 @@ import net.jodah.expiringmap.ExpiringMap;
  * Cache transforms functionality only works if <code>url</code> is used. Caching is not supported with <code>mappingSource</code>.
  * </p>
  * <p>
+ * File path context when using XSL import functionality only works if <code>url</code> is used. Caching is not supported with <code>mappingSource</code>.
+ * </p>
+ * <p>
  * If you wish to call an external mapping source when using <code>mappingSource</code> such as via http you can use
- * <strong>FileDataInputParameter</strong>.
+ * <strong>FileDataInputParameter</strong>. However that will not keep the file path context when using XSL import.
  * </p>
  *
  * @config xml-transform-service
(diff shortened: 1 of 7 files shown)


higgyfella (approved):
looks good

PR #1708, opened 2024-11-18
Implement MessageUniqueIdService

telus-terchu:
Implement config to generate new unique id for CloneMessageServiceList

## Motivation

There is a requirement to generate a new message id and set it on the message.

## Modification

- The way of obtaining the id generator from the message's factory is by calling uniqueIdGenerator(), however this needed to be changed to public access instead of protected. The logic to generate message unique id is in MessageHelper.
- Did not know what impact to things that depend on the message id such as events, or whether there is a need to correlate the new message ids with the original are needed?

## PR Checklist

- [x] been self-reviewed.

## Result
- Users will be able to configure a MessageUniqueIdService which generates a new message id.
- Users will be able to configure CloneMessageServiceList with 'newUniqueIdPerMessage' property to control whether a new message id is generated per cloned message.

## Testing
- Run the tests for MessageUniqueIdServiceTest and CloneMessageServiceListTest

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/CloneMessageServiceList.java
+++ b/interlok-core/src/main/java/com/adaptris/core/CloneMessageServiceList.java
@@ -16,20 +16,25 @@
 
 package com.adaptris.core;
 
-import static com.adaptris.core.util.LoggingHelper.friendlyName;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.stream.Collectors;
-import javax.validation.Valid;
-import org.apache.commons.lang3.ObjectUtils;
 import com.adaptris.annotation.AdapterComponent;
 import com.adaptris.annotation.AdvancedConfig;
 import com.adaptris.annotation.ComponentProfile;
 import com.adaptris.annotation.DisplayOrder;
 import com.adaptris.core.metadata.MetadataFilter;
 import com.adaptris.core.metadata.RemoveAllMetadataFilter;
+import com.adaptris.core.util.MessageHelper;
 import com.thoughtworks.xstream.annotations.XStreamAlias;
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.commons.lang3.ObjectUtils;
+
+import javax.validation.Valid;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+import static com.adaptris.core.util.LoggingHelper.friendlyName;
 
 /**
  * Implementation of {@linkplain ServiceCollection} that creates a new clone of {@linkplain com.adaptris.core.AdaptrisMessage} for each configured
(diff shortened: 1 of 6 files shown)


interlok-core/src/main/java/com/adaptris/core/CloneMessageServiceList.java
```diff
@@ -61,6 +66,10 @@ public class CloneMessageServiceList extends ServiceListBase {
   @Valid
   private MetadataFilter overrideMetadataFilter;
 
+  @Getter
```
sebastien-belin-adp, line 73: A small javadoc explanation on what it is would be nice here.
Also you could use @InputFieldDefault(value = "false") so the UI is aware of the default value and can display that to users.

interlok-core/src/test/java/com/adaptris/core/CloneMessageServiceListTest.java
```diff
@@ -87,6 +89,44 @@ public void testNormalOperation() throws Exception {
...
+
+    CloneMessageServiceList service2 = createServiceList();
+    service2.setNewUniqueIdPerMessage(true);
+    service2.getServices().add(functionService);
+    service2.getServices().add(functionService);
+
+    try {
+      LifecycleHelper.initAndStart(service1);
+      service1.doService(msg);
+      assertTrue(messageUniqueIds.size() == 1 && messageUniqueIds.contains(msg.getUniqueId()));
```
sebastien-belin-adp: A bit hard to read assert I think

interlok-core/src/main/java/com/adaptris/core/services/MessageUniqueIdService.java
```diff
@@ -0,0 +1,34 @@
...
+import com.adaptris.core.AdaptrisMessage;
+import com.adaptris.core.CoreException;
+import com.adaptris.core.ServiceException;
+import com.adaptris.core.ServiceImp;
+import com.adaptris.core.util.MessageHelper;
+
+/**
+ * This service generates a new message unique id and sets it on the message
+ */
+public class MessageUniqueIdService extends ServiceImp {
```
sebastien-belin-adp, line 25: For this one, I was probably not clear with my Friday last minute email.
We need something where users can configure which unique Id they want.
So roughly something like:

```xml
<message-unique-id-service>
  <message-unique-id>NewMessageId</message-unique-id>
</message-unique-id-service>
```

And also we should be able to use expression to get the value from metadata so something like that:
```xml
<message-unique-id-service>
  <message-unique-id>%message{metadataWithMessageId}</message-unique-id>
</message-unique-id-service>
```
Let me know if you need help with that.

And finally, have a look at https://interlok.adaptris.net/interlok-docs/#/pages/developer/developer-annotations to annotate the service. to work nicely in the UI.

sebastien-belin-adp (changes requested):

sebastien-belin-adp:
@telus-terchu 

> "Did not know what impact to things that depend on the message id such as events, or whether there is a need to correlate the new message ids with the original are needed?"

We don't need to care about this as this service was asked for some specific use case so they will know how to handle it but otherwise it's not recommended to change the message id indeed.

In fact if you could add a comment on the javadoc of this service saying that we don't recommend changing the message id and using this service should be use with care and should be used at the beginning of a workflow.

Thanks

sebastien-belin-adp (approved):
Thanks

PR #1713, opened 2024-11-21
INTERLOK-4462: MetadataAppenderService - Add optional separator

telus-terchu:
## Motivation

A requirement for adding an optional separator to MetadataAppenderService. See https://adaptris.atlassian.net/browse/INTERLOK-4462

## Modification

- MetadataAppenderService: Added a String separator property
- MetadataAppenderService: Updated the existing doService() to use stream api and join
- MetadataAppenderServiceTest: added testSeparator(), testZeroKeys()
- build.gradle: Added junit-jupiter-params test dependency to keep DRY

## PR Checklist

- [x] been self-reviewed.
- [x] Checked that new 3rd party dependencies are appropriately licensed

## Result

- Let say I have k1=v1, k2=v2 and I append k1 and k2 to k3, the result will be k3=v1v2

## Testing

- Run MetadataAppenderServiceTest.testSeparator()

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/metadata/MetadataAppenderService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/metadata/MetadataAppenderService.java
@@ -28,6 +28,7 @@ import com.adaptris.annotation.AffectsMetadata;
 import com.adaptris.annotation.AutoPopulated;
 import com.adaptris.annotation.ComponentProfile;
 import com.adaptris.annotation.DisplayOrder;
+import com.adaptris.annotation.InputFieldDefault;
 import com.adaptris.core.AdaptrisMessage;
 import com.adaptris.core.MetadataElement;
 import com.adaptris.core.util.Args;
@@ -65,6 +66,7 @@ public class MetadataAppenderService extends MetadataServiceImpl {
   @NotNull
   @AutoPopulated
   @AffectsMetadata
+  @InputFieldDefault("")
   @Getter
   @Setter
   private String separator = DEFAULT_SEPARATOR;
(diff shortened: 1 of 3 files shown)


interlok-core/src/main/java/com/adaptris/core/services/metadata/MetadataAppenderService.java
```diff
@@ -139,4 +152,11 @@ public void setResultKey(String string) {
     resultKey = Args.notBlank(string, "resultKey");
   }
 
+
+  /**
+   * Sets the separator to be used between concatenated metadata values
+   * @param string the separator between concatenated metadata values, may not be null
+   */
+  public void setSeparator(String string) { separator = Args.notNull(string, "separator"); }
```
sebastien-belin-adp: We already have lombok @Setter

interlok-core/src/main/java/com/adaptris/core/services/metadata/MetadataAppenderService.java
```diff
@@ -56,6 +62,15 @@ public class MetadataAppenderService extends MetadataServiceImpl {
...
   private String resultKey;
 
+  public static final String DEFAULT_SEPARATOR = "";
+  @NotNull
+  @AutoPopulated
+  @AffectsMetadata
+  @InputFieldDefault("")
+  @Getter
+  @Setter
+  private String separator = DEFAULT_SEPARATOR;
```
sebastien-belin-adp, line 73: Maybe we should remove @NoNull and also have a method that will return the default empty string if null.
Or use @InputFieldHint(style = "BLANKABLE") as otherwise in the UI empty value will be consider as null and the user will not be able to save this config.

sebastien-belin-adp (changes requested):

interlok-core/src/main/java/com/adaptris/core/services/metadata/MetadataAppenderService.java
telus-terchu: @sebastien-belin-adp 
I made the separator nullable and return the default separator if null. Also made it blankable.

sebastien-belin-adp (approved):
Thanks

PR #1724, opened 2024-11-27
Feature/interlok 4453 update change char encoding service expressions

saurav-mukherjee-telus:
## Motivation

Why am I doing this

## Modification

What did I change

## PR Checklist

- [x] been self-reviewed.

## Result

## Testing

### DIFF
--- a/interlok-core/src/test/java/com/adaptris/core/services/ChangeCharEncodingServiceTest.java
+++ b/interlok-core/src/test/java/com/adaptris/core/services/ChangeCharEncodingServiceTest.java
@@ -59,10 +59,11 @@ public class ChangeCharEncodingServiceTest extends GeneralServiceExample {
   public void testSetCharEncodingExpression() throws Exception {
     ChangeCharEncodingService srv = new ChangeCharEncodingService();
     assertNull(srv.getCharEncoding());
-    srv.setCharEncoding("%message{pn.processing.worker.transform.encoding-target}");
     AdaptrisMessage msg = AdaptrisMessageFactory.getDefaultInstance().newMessage("Hello");
-    assertEquals(null, msg.getCharEncoding());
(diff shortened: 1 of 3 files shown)


interlok-core/src/test/java/com/adaptris/core/services/ChangeCharEncodingServiceTest.java
```diff
@@ -55,4 +55,19 @@ public void testChangeCharset() throws Exception {
     assertEquals(Charset.forName("iso-8859-1"), Charset.forName(msg.getContentEncoding()));
   }
 
+  @Test
+  public void testSetCharEncodingExpression() throws Exception {
+    ChangeCharEncodingService srv = new ChangeCharEncodingService();
+    assertNull(srv.getCharEncoding());
+    srv.setCharEncoding("%message{pn.processing.worker.transform.encoding-target}");
```
sebastien-belin-adp: You don't/shouldn't need to use pn metadata here as Interlok is not PN.
But it doesn't matter.

sebastien-belin-adp (approved):

sebastien-belin-adp (changes requested):
If you want to improve the coverage you could add another test with an empty char encoding value. That will test one more branch of the code.

interlok-core/src/test/java/com/adaptris/core/services/ChangeCharEncodingServiceTest.java
saurav-mukherjee-telus: Thanks @sebastien-belin-adp

saurav-mukherjee-telus:
> If you want to improve the coverage you could add another test with an empty char encoding value. That will test one more branch of the code.

Hi @sebastien-belin-adp  - Null character value test is in place now

sebastien-belin-adp (approved):
Thanks

PR #1726, opened 2024-11-29
Adding optional separator for Metadata Appender Service.

saurav-mukherjee-telus:
## Motivation

Why am I doing this

## Modification

What did I change

## PR Checklist

- [x] been self-reviewed.

## Result

## Testing

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/metadata/MetadataAppenderService.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/metadata/MetadataAppenderService.java
@@ -20,16 +20,16 @@ import java.util.ArrayList;
 import java.util.List;
 import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
-import com.adaptris.annotation.AdapterComponent;
-import com.adaptris.annotation.AffectsMetadata;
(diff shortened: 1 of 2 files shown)


sebastien-belin-adp:
Hi @saurav-mukherjee-telus when you create a PR can you please fill up all the fields in the descriptions.
It's probably not very important for this task but it's a good habit to have.
You don't need to write a novel but a couple of lines / bullet points that help us understand the PR.
And also providing a way to test your changes.
An example xml config can be a good thing.

Then for the changes, I was saying that to Terry as well.
It's good to use the Interlok annotations described in https://interlok.adaptris.net/interlok-docs/#/pages/developer/developer-annotations such as   @AffectsMetadata, @InputFieldHint(style="BLANKABLE"), @InputFieldDefault("")
So the UI can display the right information or behave accordingly.
In this case @InputFieldHint(style="BLANKABLE") is necessary to tell the UI separator can be an empty value instead of a null value.
And also, there is a small issue, think what will happen if we set a null value to separator.

Thanks

saurav-mukherjee-telus:
Closing the PR  since the changes are already addressed in another PR - https://github.com/adaptris/interlok/pull/1713

PR #1736, opened 2024-12-10
INTERLOK-4450 Add interlok-jetty-configurable-key-security-handler into Interlok core

sebastien-belin-adp:
## Motivation

The custom components were used a lot but hard to maintain as not part of the Interlok distribution

## Modification

- Add ConfigurableKeySecurityHandler and its config classes
- Add tests

## PR Checklist

- [x] been self-reviewed.

## Result

There is a new jetty security handler that can handle keys

## Testing

That has already been tested.
The unit tests should pass.


higgyfella (approved):
looks good

PR #1737, opened 2024-12-12
INTERLOK-3834: Handle most file paths

telus-terchu:
## Motivation

https://adaptris.atlassian.net/browse/INTERLOK-3834

## Modification

Mostly FsHelper.java where the logic update was, other classes comments were changed as per https://github.com/adaptris/interlok/pull/1194

## PR Checklist

- [x] been self-reviewed.

## Result

Support URLs such as:
"exceptions/duplicate"
"./exceptions/duplicate"
"../exceptions/duplicate"
"file://localhost/./fred"
"file://localhost/./fred/wilma"
"file:./a/b.xml"
"file://./a1/b1.xml"
"file://../a2/b2.xml"
"file:///./a3/b3.xml"
"file:///a4/b4.xml"
"file:////a5/b5.xml"

## Testing

Run FsHelperTest.java

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/fs/FsHelper.java
+++ b/interlok-core/src/main/java/com/adaptris/core/fs/FsHelper.java
@@ -42,6 +42,7 @@ import com.adaptris.interlok.util.FileFilterBuilder;
 public abstract class FsHelper {
 
   private static transient Logger log = LoggerFactory.getLogger(FsHelper.class);
+  private static final String WIN_DRIVE_REGEX = "^[a-zA-Z]{1}$"; //matches an alphabetic character exactly 1 time in the string.
 
   /**
    * Go straight to a {@link File} from a url style string.
@@ -53,6 +54,8 @@ public abstract class FsHelper {
       return createFileReference(createUrlFromString(s, true));
     } catch (IllegalArgumentException e) {
       // Catch it from createUrlFromString(), since it's probably c:/file.
+      // strip out any uri scheme
+      s = s.replaceFirst("file://", "");
       return new File(s);
     }
   }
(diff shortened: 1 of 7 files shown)


interlok-core/src/main/java/com/adaptris/core/fs/FsHelper.java
```diff
@@ -53,6 +54,8 @@ public static File toFile(String s) throws IOException, URISyntaxException {
       return createFileReference(createUrlFromString(s, true));
     } catch (IllegalArgumentException e) {
       // Catch it from createUrlFromString(), since it's probably c:/file.
+      // strip out any uri scheme
+      s = s.replaceFirst("file://", "");
```
sebastien-belin-adp: @telus-terchu This line is not covered by tests but seems to be important to cope with windows path.

sebastien-belin-adp (changes requested):

interlok-core/src/main/java/com/adaptris/core/fs/FsHelper.java
telus-terchu: @sebastien-belin-adp I removed that s = s.replaceFirst() line I think its not needed.
Also, based on the updated code, the only case an IllegalArgumentException would be thrown if it was not a "file" scheme, e.g. http://localhost/./fred. While the current logic would return a valid File object, should it be supported or should it throw an exception?

sebastien-belin-adp: @telus-terchu I didn't really think about that but I guess it could be supported. I can imagine some places where you could use a remote http file. Not in the FsConsumer but in other place which would read something from a file.

telus-terchu: @sebastien-belin-adp I've left the existing logic as is. The FsHelper.createUrlFromString() will throw an IllegalArgumentException if it encounters a non "file" scheme, while the FsHelper.toFile() will catch that exception and return a new File() regardless if it is valid. This is because the tests fail if the IllegalArgumentException is not thrown.

sebastien-belin-adp (approved):

PR #1747, opened 2024-12-19
Message Splitter which keep parent node

telus-terchu:
## Motivation

https://adaptris.atlassian.net/browse/INTERLOK-4469

## Modification

Added a flag on XpathMessageSplitter called "retainBranchNodes" which will retain the split node's parents and their siblings right up to the root node.

## PR Checklist

- [x] been self-reviewed.

## Result
For a document:
<?xml version ="1.0" encoding="UTF-8"?>
            <envelope>
              <header>
                <po>123123</po>
              </header>
              <body>
                <document>one</document>
                <document>two</document>
                <document>three</document>
              </body>
              <trailer>
                <x>123</x>
              </trailer>
            </envelope>

When the flag = true, the split will end up as:

<?xml version ="1.0" encoding="UTF-8"?>
            <envelope>
              <header>
                <po>123123</po>
              </header>
              <body>
                <document>one</document>
              </body>
              <trailer>
                <x>123</x>
              </trailer>
            </envelope>

<?xml version ="1.0" encoding="UTF-8"?>
            <envelope>
              <header>
                <po>123123</po>
              </header>
              <body>
                <document>two</document>
              </body>
              <trailer>
                <x>123</x>
              </trailer>
            </envelope>

<?xml version ="1.0" encoding="UTF-8"?>
            <envelope>
              <header>
                <po>123123</po>
              </header>
              <body>
                <document>three</document>
              </body>
              <trailer>
                <x>123</x>
              </trailer>
            </envelope>

## Testing

See XpathSplitterTest.testSplit()

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/services/splitter/XpathMessageSplitter.java
+++ b/interlok-core/src/main/java/com/adaptris/core/services/splitter/XpathMessageSplitter.java
@@ -18,6 +18,9 @@ package com.adaptris.core.services.splitter;
 
 import static com.adaptris.core.util.XmlHelper.createDocument;
 import java.io.IOException;
+import java.util.Collections;
+import java.util.LinkedList;
+import java.util.List;
 import javax.validation.constraints.NotBlank;
 import javax.xml.namespace.NamespaceContext;
 import javax.xml.parsers.DocumentBuilder;
@@ -25,6 +28,7 @@ import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 import javax.xml.xpath.XPathExpressionException;
 import org.w3c.dom.Document;
+import org.w3c.dom.Element;
 import org.w3c.dom.Node;
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
@@ -74,7 +78,7 @@ import com.thoughtworks.xstream.annotations.XStreamAlias;
  * @author sellidge
  */
 @XStreamAlias("xpath-message-splitter")
-@DisplayOrder(order = {"xpath", "encoding", "copyMetadata", "copyObjectMetadata", "namespaceContext", "xmlDocumentFactoryConfig"})
+@DisplayOrder(order = {"xpath", "encoding", "copyMetadata", "copyObjectMetadata", "retainBranchNodes", "namespaceContext", "xmlDocumentFactoryConfig"})
 public class XpathMessageSplitter extends MessageSplitterImp {
 
   @NotBlank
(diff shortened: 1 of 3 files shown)


interlok-core/src/main/java/com/adaptris/core/services/splitter/XpathMessageSplitter.java
```diff
@@ -217,13 +235,58 @@ private void initialiseSplitGenerator() throws Exception {
...
+      }
+      return root;
+    }
+
     @Override
     protected AdaptrisMessage constructAdaptrisMessage() throws Exception {
       if (nodeListIndex < nodeList.getLength()) {
         Node e = nodeList.item(nodeListIndex);
         Document splitXmlDoc = docBuilder.newDocument();
         Node dup = splitXmlDoc.importNode(e, true);
```
sebastien-belin-adp: Do we still need that line Node dup = splitXmlDoc.importNode(e, true);?

interlok-core/src/main/java/com/adaptris/core/services/splitter/XpathMessageSplitter.java
```diff
@@ -186,6 +192,18 @@ public void setXmlDocumentFactoryConfig(DocumentBuilderFactoryBuilder xml) {
     this.xmlDocumentFactoryConfig = xml;
   }
 
+  public boolean isRetainBranchNodes() {
```
sebastien-belin-adp: You need to use a getter, the 'is' is not working well with the UI.

interlok-core/src/test/java/com/adaptris/core/services/splitter/XpathSplitterTest.java
```diff
@@ -167,18 +167,22 @@ public void testSplit() throws Exception {
...
+    XpathMessageSplitter splitter = new XpathMessageSplitter("/envelope/body/document");
+    splitter.setRetainBranchNodes(true);
+    XmlUtils xml = new XmlUtils();
+    int count = 0;
     try (CloseableIterable<AdaptrisMessage> closeable = splitter.splitMessage(msg)) {
       for (AdaptrisMessage m : closeable) {
         assertFalse(m.getObjectHeaders().containsKey(obj));
         xml.setSource(new ByteArrayInputStream(m.getPayload()));
-        assertNotNull(xml.getSingleNode("/envelope/body/document"));
+        assertEquals(xml.getSingleNode("/envelope/body/document/text()").getNodeValue(), String.valueOf(count+1));
```
sebastien-belin-adp, line 186: It assert that the node exist but maybe we should check that the other nodes don't exist?

sebastien-belin-adp (changes requested):

sebastien-belin-adp (approved):

PR #1979, opened 2025-07-31
updated test resources to use dummy values

higgyfella:
## Modification

changed the values in the test resources that tested the password encoding.


juanromero86 (approved):
Thanks Paul.

PR #1998, opened 2025-08-18
implemented new RetryStore method to get error stacktrace

jwilliamson15:
Implementing a new getStacktrace method within RetryStore. This enables 2 new endpoints from RetryFromJetty - "/api/failed/stacktrace/{id}" 
"/api/failed/stacktrace/first-line/{id}"

These endpoints will return the stacktrace or first line given a message id.

### DIFF
--- a/interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/FilesystemRetryStore.java
+++ b/interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/FilesystemRetryStore.java
@@ -8,6 +8,7 @@ import java.io.FileWriter;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintWriter;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -249,4 +250,17 @@ public class FilesystemRetryStore implements RetryStore {
   public void makeConnection(AdaptrisConnection connection) {
     // null implementation 
   }
+
+    @Override
+    public String getStackTrace(String msgId) throws InterlokException {
+        try {
+            File dir = validateMsgId(msgId, true);
+            File stackTraceFile = new File(dir, STACKTRACE_FILENAME);
+            FsWorker.checkReadable(FsWorker.isFile(stackTraceFile));
+
+            return FileUtils.readFileToString(stackTraceFile, StandardCharsets.UTF_8);
+        } catch (Exception e) {
+            throw ExceptionHelper.wrapInterlokException(e);
+        }
+    }
 }
--- a/interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/FilesystemRetryStore.java
+++ b/interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/FilesystemRetryStore.java
@@ -136,10 +136,7 @@ public class FilesystemRetryStore implements RetryStore {
     return validateDir(target, mustAlreadyExist);
   }
 
-  /**
-   * Validate that the given string is a safe path component (no separators, no "..", not absolute).
-   */
-  private static void validatePathComponent(String component) {
+  static void validatePathComponent(String component) {
     if (component == null || component.isEmpty()) {
       throw new IllegalArgumentException("Message ID may not be null or empty");
     }
--- a/interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/FilesystemRetryStore.java
+++ b/interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/FilesystemRetryStore.java
@@ -131,8 +131,21 @@ public class FilesystemRetryStore implements RetryStore {
   }
 
   private File validateMsgId(String msgId, boolean mustAlreadyExist) throws Exception {
+    validatePathComponent(msgId);
     File target = new File(FsHelper.toFile(getBaseUrl()), msgId);
     return validateDir(target, mustAlreadyExist);
+  }
+
+  /**
+   * Validate that the given string is a safe path component (no separators, no "..", not absolute).
+   */
+  private static void validatePathComponent(String component) {
+    if (component == null || component.isEmpty()) {
+      throw new IllegalArgumentException("Message ID may not be null or empty");
+    }
+    if (component.contains("..") || component.contains("/") || component.contains("\\") || new File(component).isAbsolute()) {
+      throw new IllegalArgumentException("Invalid message ID: path traversal or separator detected");
+    }
   }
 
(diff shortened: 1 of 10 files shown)


interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/RetryFromJetty.java
```diff
@@ -57,433 +50,598 @@
...
+
+        protected void handleException(Exception e, AdaptrisMessage msg) {
+            msg.setContent(ExceptionUtils.getRootCauseMessage(e), StandardCharsets.UTF_8.name());
+            sendResponse(HTTP_ERROR, msg);
+        }
+
+        protected void handleStackTraceResponse(String msgId, AdaptrisMessage jettyMsg, String stackTrace) {
+            String httpCode = (msgId != null) ? HTTP_OK : HTTP_BAD;
+            if (msgId != null) {
+                jettyMsg.setContent(stackTrace, StandardCharsets.UTF_8.name());
```
telus-terchu, line 460: - What should we send back if stackTrace is null?
- AdaptrisMessage.setContent() seems to be removed in 5.1, should we use setPayload()?

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/RetryFromJetty.java
```diff
@@ -57,433 +50,598 @@
...
+                    if (deleted) {
+                        httpCode = HTTP_OK;
+                    } else {
+                        httpCode = HTTP_NOT_FOUND;
+                    }
+                } else {
+                    httpCode = HTTP_BAD;
+                }
+            } catch (Exception e) {
+                handleException(e, jettyMsg);
```
telus-terchu, line 539: Should the method return after catching the exception? Otherwise the sendResponse() will be called with HTTP_ERROR code

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/RetryFromJetty.java
```diff
@@ -57,433 +50,598 @@
...
+                        public void run() {
+                            Thread.currentThread().setName("Retry Failed Message");
+                            workflow.onAdaptrisMessage(msgForRetry, success, failure);
+                        }
+                    });
+                } else {
+                    httpCode = HTTP_BAD;
+                }
+            } catch (Exception e) {
+                handleException(e, jettyMsg);
```
telus-terchu, line 590: Should the method return after catching the exception? Otherwise the sendResponse() will be called with HTTP_ERROR code

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/RetryFromJetty.java
```diff
@@ -57,433 +50,598 @@
...
+                    // Do we want people to configure it?
+                    // Therefore we look up the metadata from the store;
+                    // Figure out the workflow, and then get the consumer.getMessageFactory()
+                    Map<String, String> metadata = retryStore.getMetadata(msgId);
+                    Workflow workflow = getWorkflow(metadata.get(Workflow.WORKFLOW_ID_KEY));
+                    AdaptrisMessage msgForRetry =
+                            retryStore.buildForRetry(msgId, metadata, workflow.getConsumer().getMessageFactory());
+                    // We know at this point we have something to retry.
+                    // So, we can fire a 202 before submission.
+                    httpCode = HTTP_ACCEPTED;
```
telus-terchu, line 572: There is a logic difference between existing implementation, which calls sendReponse() at this point, compared to at the end of the method probably doesn't make a huge difference but just wanted to highlight it.

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/FilesystemRetryStore.java
```diff
@@ -130,10 +131,20 @@ private void storePayload(AdaptrisMessage msg, File payloadFile) throws Exceptio
...
 
   private File validateMsgId(String msgId, boolean mustAlreadyExist) throws Exception {
+    validatePathComponent(msgId);
     File target = new File(FsHelper.toFile(getBaseUrl()), msgId);
     return validateDir(target, mustAlreadyExist);
   }
 
+  static void validatePathComponent(String component) {
+    if (component == null || component.isEmpty()) {
+      throw new IllegalArgumentException("Message ID may not be null or empty");
```
saurav-mukherjee-telus, line 141: @jwilliamson15 Please check if exception should read as - throw new IllegalArgumentException("Message ID may be null or empty")?, Thank you

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/RetryFromJetty.java
jwilliamson15: Good question for the stacktrace, I think we just return null too as that is what is in the store although this would be an unlikely scenario. I will double check though.

As for setContent(), I think it is still in use, I can't see anything to say it is deprecated.

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/FilesystemRetryStore.java
jwilliamson15: This is a null check and then we throw the error because msgId must contain a path with the FilesystemRetryStore such as /home/josh/myfile

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/RetryFromJetty.java
jwilliamson15: This is by design as an exception would return a http 500 error code. That is returned in the handleException method

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/RetryFromJetty.java
jwilliamson15: good spot! I will fix this

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/RetryFromJetty.java
jwilliamson15: same as above

interlok-core/src/main/java/com/adaptris/core/http/jetty/retry/RetryFromJetty.java
telus-terchu: Oh yes you're right, setStringPayload() is deprecated

PR #2073, opened 2026-08-06
Retry store write: backfill workflowId before persisting failed messages to S3

higgyfella:
This patch makes retry-store writes operationally safer by ensuring workflowId is present in failed-message metadata before writing to S3.

Problem:
- Some failed messages are being persisted without workflowId.
- Retry APIs then fail with No Workflow [null] found because workflow resolution depends on workflowId in metadata.properties.

What changed:
- Updated RetryStoreWriteService.java to enrich workflowId before calling retryStore.write(msg).
- Enrichment logic is:
1. Keep existing workflowId if already present.
2. If missing, use MessageLifecycleEvent.workflowId.
3. If still missing, check legacy lowercase workflowid metadata key.
4. If found, write it back as workflowId.

- Added RetryStoreWriteTest.java test to verify workflowId is backfilled from MessageLifecycleEvent and persisted to metadata.properties.

### DIFF
--- a/gradle/owasp-exclude.xml
+++ b/gradle/owasp-exclude.xml
@@ -391,4 +391,14 @@
     <packageUrl regex="true">^pkg:maven/org\.eclipse\.jetty(\.http2|\.websocket)?/.+@10\.0\.26$</packageUrl>
     <cve>CVE-2026-2332</cve>
   </suppress>
+  <suppress>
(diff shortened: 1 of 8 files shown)


jwilliamson15 (approved):
Tested and confirmed the workflowId is present in metadata.properties after this fix.

